Issue #4, opened 2022-07-03
[Feature]: Implement basic Milvus client in Rust


yah01:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

For now, a basic Milvus client requires:

- [x] Implement Connect
- [x] Implement CreateCollection, LoadCollection, ReleaseCollection and DropCollection
- [x] Implement Search and Query
- [x] Implement CreateIndex

PR #5, opened 2022-07-03
Add client connection for implement basic Milvus client

HolmiumTS:
#4 
add client connection and corresponding unit tests

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -14,4 +14,24 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub struct Client {}
+use crate::error::{Error, Result};
+use crate::proto::milvus::milvus_service_client::MilvusServiceClient;
+use tonic::codegen::StdError;
+use tonic::transport::Channel;
+
+pub struct Client {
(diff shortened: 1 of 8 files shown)


src/client.rs
```diff
@@ -14,4 +14,55 @@
...
+use crate::proto::milvus::milvus_service_client::MilvusServiceClient;
+use tonic::codegen::StdError;
+use tonic::transport::Channel;
+
+pub struct Client {
+    client: MilvusServiceClient<Channel>,
+}
+
+impl Client {
+    pub async fn new<D>(dst: D) -> Result<Self>
```
yah01, line 27: Does `tonic::transport::Endpoint` require the protocol prefix like "http://"? If yes, the parameter(s) should be `(addr: &str)`, or `(host: &str, port: i32)`, which hide the unimportant details (prefer the former)

yah01 (changes requested):

src/client.rs
yah01: or Into<String>

PR #11, opened 2022-08-10
Add insert, flush, and query methods to collection

andreytkachenko:
Here are the proposing changes
* `Client::flush_collections`
* `Collection::create_insert_frame` (temporal solution) I think it would be better to store schema inside collection itself
* `CollectionFieldsData` and `FieldData` 
* `Collection::insert`
* `Collection::query`

### DIFF
--- a/examples/collection.rs
+++ b/examples/collection.rs
@@ -1,0 +1,56 @@
+use milvus::{client::Client, error::Error, schema::AsFieldDataValue};
+
+// #[derive(Debug, Clone, milvus::Entity)]
+// struct ImageEntity {
+//     #[milvus(primary, auto_id = false)]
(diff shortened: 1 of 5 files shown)


yah01:
Hi @andreytkachenko,
Thanks for your contribution, some questions for your proposals:
- the example requires the collection existed, create it if not existed would be better
- create_insert_frame is nice, the users won't create wrong data schema with it, could you add a method add_batch(), which is column-oriented? It's useful for performance-sensitive cases (the parameter may be: the item of iterator is (field name,batch of data)).

yah01:
Hi @andreytkachenko,
I decide to merge this PR, because the coming PRs of others are based on this, thanks for your contribution!

PR #12, opened 2022-08-17
[WIP] Entity based collections

andreytkachenko:
Here is my final (almost) vision of how collections should be based on entities
key notes:
- `Entity` - is main trait describing schema of the collection
- `Collection` - is a trait describing batch
- Each collection is constrained by appropriate entity type

what is not finished:
- procedural macros 

Current Example:
```rust 
struct Image {
    id: i64,
    hash: Vec<u8>,
   ...
}
impl Entity for Image {
    const SCHEMA = [
         ...
    ];
    ...
}
struct ImageBatch {
    id: Vec<i64>,
    hash: Vec<u8>,
   ...
}
impl Collection for ImageBatch {
    type Entity = Image;
    ....
}
```

Final (with proc_macro) example:
```rust
#[derive(Debug, Clone, Default, milvus::Entity)]
struct Image {
    #[milvus(primary, auto)]
    pub id: i64,

    #[milvus(binary)]
    pub hash: Vec<u8>,
    
    #[milvus(name = "listing_id")]
    pub listingId: i32,

    ...
}
// ImageBatch will be generated aswell
```


yah01:
Please check [these errors](https://github.com/milvus-io/milvus-sdk-rust/runs/7883127710?check_suite_focus=true#step:4:339)

PR #21, opened 2022-10-09
Simplify search() method

yah01:
Signed-off-by: yah01 <yang.cen@zilliz.com>


kingzcheung:
I think some structs should implement #[derive(Debug, Default)]

yah01:
> I think some structs should implement #[derive(Debug, Default)]

Thanks for feedback, could you create an issue about this? I just implemented `Default` for `SearchOption`, any lacks?

PR #24, opened 2022-10-13
Refactor index related methods

yah01:
Signed-off-by: yah01 <yang.cen@zilliz.com>


yah01:
This PR prepares for Milvus v2.2

PR #80, opened 2024-01-11
impl delete back

yah01:
/kind improvement


src/mutate.rs
```diff
@@ -102,8 +118,92 @@ impl Client {
...
+                options.filter.len().to_string(),
+            ));
+        }
+
+        if options.ids.len() == 0 && options.filter.len() == 0 {
+            return Err(Error::InvalidParameter(
+                "len(filter)".to_owned(),
+                options.filter.len().to_string(),
+            ));
+        }
```
sunby, line 144: better to check in one if statement

src/mutate.rs
```diff
@@ -102,8 +118,92 @@ impl Client {
...
+                    }
+
+                    _ => {
+                        return Err(Error::InvalidParameter(
+                            "pk type".to_owned(),
+                            pk.dtype.as_str_name().to_owned(),
+                        ));
+                    }
+                }
+            }
```
sunby, line 184: prefer to organize these codes to a member function (eg. makeExpr) to make it clearer

src/mutate.rs
```diff
@@ -102,8 +118,92 @@ impl Client {
...
+                "len(filter)".to_owned(),
+                options.filter.len().to_string(),
+            ));
+        }
+
+        let collection_name = collection_name.into();
+
+        let mut expr = match options.filter.len() {
+            0 => {
+                let collection = self.collection_cache.get(&collection_name).await?;
```
sunby, line 150: what would happen if another sdk is modifing this collection in another process?

src/mutate.rs
```diff
@@ -102,8 +118,92 @@ impl Client {
...
+
+        Ok(result)
+    }
+
+    pub async fn delete(
+        &self,
+        collection_name: impl Into<String>,
+        options: &DeleteOptions,
+    ) -> Result<crate::proto::milvus::MutationResult> {
+        if options.ids.len() > 0 && options.filter.len() > 0 {
```
sunby, line 132: maybe we can do something when constrcuting DeleteOptions to avoid set both of `ids` and `filter`? like only offer `with_ids` and `with_filter` to set these fields.

PR #85, opened 2024-01-17
add range_search test

LeoReeYang:
append
- add test `collection_range_search` for range search

### DIFF
--- a/tests/collection.rs
+++ b/tests/collection.rs
@@ -189,3 +189,57 @@ async fn collection_search() -> Result<()> {
     client.drop_collection(schema.name()).await?;
     Ok(())
 }
+
+#[tokio::test]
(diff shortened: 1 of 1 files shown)


tests/collection.rs
```diff
@@ -22,6 +22,8 @@ use milvus::index::{IndexParams, IndexType, MetricType};
 use milvus::mutate::InsertOptions;
 use milvus::options::LoadOptions;
 use milvus::query::{QueryOptions, SearchOptions};
+use milvus::value;
+use core::f32;
```
yah01, line 26: the two could be removed

yah01 (changes requested):

tests/collection.rs
LeoReeYang: ok

Issue #86, opened 2024-03-20
[Feature]: Index for scalar fields


heyskylark:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Hello,

I've searched through the docs and existing examples and I can't seem to find the implementation for scalar indexes. I tried using create_index on a scalar field, but it failed with `Error: Schema(NotVectorField("field_name"))`, which makes sense since it appears that IndexParams only has fields for embedding indexes.

Does the rust SDK currently have indexing for scalars? If not, is there plan to add that eventually?

Thanks!
Brandon

### Describe the solution you'd like.

A type to create scalar indexes.

yhmo:
ClientV2 already supports it: https://github.com/milvus-io/milvus-sdk-rust/pull/129

PR #88, opened 2024-05-10
Support index for scalar fields and filter describe index result with field name

aoiasd:
relate: https://github.com/milvus-io/milvus-sdk-rust/issues/86


yhmo:

PR #95, opened 2025-06-18
Update milvus rust sdk to milvus2.5

Yinwei-Yu:
This pr aims to update the milvus rust sdk to milvus 2.5.


SpadeA-Tang:
please use `cargo fmt` to format the project.

PR #103, opened 2026-02-12
feat: enable grpc keepalive with 10s interval and 5s timeout

haorenfsa:
## Summary
- Configure tonic `Endpoint` with HTTP/2 keepalive settings for faster dead connection detection
  - `http2_keep_alive_interval`: 10s (previously not set)
  - `keep_alive_timeout`: 5s (previously not set)
  - `keep_alive_while_idle`: true (previously not set)
- Add integration test verifying client connects successfully with keepalive enabled

## Test plan
- [ ] Verify `cargo check` passes (compilation)
- [ ] Verify `client_keepalive` integration test passes with a running Milvus instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -172,7 +172,11 @@ impl Client {
             Error::InvalidParameter("url".to_owned(), format!("to parse {:?}", err))
         })?;
 
-        dst = dst.timeout(timeout);
+        dst = dst
+            .timeout(timeout)
+            .http2_keep_alive_interval(Duration::from_secs(10))
+            .keep_alive_timeout(Duration::from_secs(5))
+            .keep_alive_while_idle(true);
 
         let token = match (username, password) {
             (Some(username), Some(password)) => {
--- a/tests/testsoutsideoftestsuite/client_keepalive.rs
+++ b/tests/testsoutsideoftestsuite/client_keepalive.rs
@@ -1,0 +1,11 @@
+use milvus::client::*;
+
+const URL: &str = "http://localhost:19530";
+
+#[tokio::test]
+async fn test_client_keepalive() {
+    // Verify client connects successfully with keepalive settings
+    // (10s interval, 5s timeout, enabled while idle)
+    let client = Client::new(URL).await;
+    assert!(client.is_ok(), "Client should connect with keepalive enabled");
+}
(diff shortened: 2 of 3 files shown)


src/client.rs
```diff
@@ -172,7 +172,11 @@ impl Client {
             Error::InvalidParameter("url".to_owned(), format!("to parse {:?}", err))
         })?;
 
-        dst = dst.timeout(timeout);
+        dst = dst
```
liliu-z, line 175: The three keepalive parameters (10s interval, 5s timeout, idle=true) are hardcoded in `Client::with_timeout()` which is the sole connection entry point for `Client::new()`, `ClientBuilder::build()`, and all examples. Users cannot adjust or disable these settings.

This creates real compatibility risk in proxy/LB/sidecar environments (e.g. Istio/Envoy, AWS ALB, Nginx) where HTTP/2 PING enforcement may differ from Milvus server defaults, potentially triggering GOAWAY / too_many_pings disconnections. The `keep_alive_while_idle(true)` setting also contradicts the gRPC recommendation to avoid keepalive without active calls, and Go gRPC's default `PERMIT_KEEPALIVE_WITHOUT_CALLS=false` server policy.

Suggestion: Add `keepalive_interval()`, `keepalive_timeout()`, and `keepalive_while_idle()` methods to `ClientBuilder`, retaining current values as defaults. Extract default values to `config.rs` to match the existing pattern used by `RPC_TIMEOUT`.

tests/testsoutsideoftestsuite/client_keepalive.rs
```diff
@@ -0,0 +1,11 @@
+use milvus::client::*;
+
+const URL: &str = "http://localhost:19530";
+
+#[tokio::test]
```
liliu-z, line 5: This test is functionally equivalent to the existing `client_with_timeout` test — both only assert `Client::new(URL).await.is_ok()`. It does not verify that keepalive is actually configured or that the connection survives an idle period exceeding the keepalive interval. If the keepalive configuration were removed entirely, this test would still pass.

Suggestion: If keeping this test, extend it to cover "connect → idle beyond one keepalive cycle → issue RPC" to validate the behavior this PR actually introduces.

PR #104, opened 2026-02-25
fix(collection): wait for flush completion instead of returning immediately

marlon-costa-dc:
## Summary
- The `flush` method previously returned immediately after sending the request.
- This PR updates it to poll `GetFlushState` and wait for the flush to actually complete (up to 30s timeout).
- This prevents race conditions where subsequent operations fail because the flush hasn't finished.


yhmo:

PR #105, opened 2026-02-25
fix(error): use status.code instead of error_code for Milvus > 2.3 compatibility

marlon-costa-dc:
## Summary
- Milvus server versions > 2.3 use `status.code` instead of `status.error_code`.
- This PR updates `error.rs` and `utils.rs` to map `status.code` to `ErrorCode` properly.
- Falls back to `UnexpectedError` with the original code and reason if the code is unknown.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -16,6 +16,8 @@
 // For custom error handling
 // TODO
 
+use std::convert::TryFrom;
+
 use crate::collection::Error as CollectionError;
 use crate::proto::common::{ErrorCode, Status};
 use crate::schema::Error as SchemaError;
@@ -73,7 +75,13 @@ pub enum Error {
 
 impl From<Status> for Error {
     fn from(s: Status) -> Self {
-        Error::Server(ErrorCode::from_i32(s.error_code).unwrap(), s.reason)
+        let code = ErrorCode::try_from(s.code).unwrap_or(ErrorCode::UnexpectedError);
+        let reason = if code == ErrorCode::UnexpectedError && s.code != code as i32 {
+            format!("{} (code {})", s.reason, s.code)
+        } else {
+            s.reason
+        };
+        Error::Server(code, reason)
     }
 }
 
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -14,6 +14,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::convert::TryFrom;
+
 use crate::{
     error::Error,
     proto::common::{ErrorCode, Status},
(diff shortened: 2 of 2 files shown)


src/utils.rs
```diff
@@ -24,14 +26,12 @@ pub fn status_to_result(status: &Option<Status>) -> Result<(), Error> {
...
-    match ErrorCode::from_i32(status.error_code) {
-        Some(i) => match i {
-            ErrorCode::Success => Ok(()),
-            _ => Err(Error::from(status)),
-        },
-        None => Err(Error::Unexpected(format!(
-            "unknown error code {}",
-            status.error_code
-        ))),
+    match ErrorCode::try_from(status.code) {
```
liliu-z, line 29: This PR completely switches from `status.error_code` to `status.code`, but Milvus servers < 2.3 never populate `status.code` (protobuf default = 0 = Success). Any real error from an older server will be misclassified as `Success` and silently swallowed — the exact inverse of the bug this PR fixes.

The vendored proto confirms `code` is field tag 3 (new) vs `error_code` tag 1 (deprecated but present). pymilvus handles this by checking both: `if status.code != 0 or status.error_code != 0`.

Fix: read `status.code` first; when it is 0, fall back to `status.error_code`. Apply the same logic in `error.rs` `From<Status>`.

src/error.rs
```diff
@@ -73,7 +75,13 @@ pub enum Error {
 
 impl From<Status> for Error {
     fn from(s: Status) -> Self {
-        Error::Server(ErrorCode::from_i32(s.error_code).unwrap(), s.reason)
+        let code = ErrorCode::try_from(s.code).unwrap_or(ErrorCode::UnexpectedError);
```
liliu-z, line 78: The `From<Status>` impl reads only `s.code`, so for Milvus < 2.3 servers it will produce `Error::Server(Success, …)` for genuine errors. This is the same compatibility regression as in `utils.rs` and needs the same `code`-then-`error_code` fallback.

liliu-z, line 78: Milvus >= 2.3 writes new-framework error codes into `status.code` (e.g. 100 = collection not found, 700 = index not found). These values do not match the old `ErrorCode` enum variants. `ErrorCode::try_from(status.code)` therefore maps to the wrong variant or falls through to `UnexpectedError`, losing the specific error type. Impact is limited because `reason` still carries the correct text, but any user code that pattern-matches on `ErrorCode` will hit the wrong arm.

src/utils.rs
```diff
@@ -24,14 +26,12 @@ pub fn status_to_result(status: &Option<Status>) -> Result<(), Error> {
...
+        Ok(ErrorCode::Success) => Ok(()),
```
liliu-z, line 30: `status_to_result()` and `From<Status> for Error` are the global error gateway for every RPC call, yet there are no unit tests covering them. At minimum, tests should cover: `code = 0` (success), `code = known non-zero`, `code = unknown value` (try_from failure), and the compatibility branch `code = 0 && error_code != 0`.

src/utils.rs
```diff
@@ -24,14 +26,12 @@ pub fn status_to_result(status: &Option<Status>) -> Result<(), Error> {
...
+        Ok(_) => Err(Error::from(status)),
+        Err(_) => Err(Error::Server(
```
liliu-z, line 32: The `Err(_)` branch in `status_to_result()` (lines 32-35) manually constructs `Error::Server(UnexpectedError, …)`, duplicating the same fallback already in `From<Status>` (error.rs lines 78-83). This creates two independent maintenance points. The `Err(_)` branch can simply delegate to `Err(Error::from(status))`.

yhmo:

PR #106, opened 2026-02-25
fix(iterator): fallback to iterator consistency level if not provided in options

marlon-costa-dc:
## Summary
- When creating a `QueryIterator` or `SearchIterator`, if `consistency_level` is not explicitly provided in the options, it now falls back to the iterator's own `consistency_level` instead of hardcoding `0`.

### DIFF
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -554,7 +554,10 @@ impl QueryIterator {
                 guarantee_timestamp: 0,
                 query_params: init_ts_params,
                 not_return_all_meta: false,
-                consistency_level: self.options.consistency_level.unwrap_or(0),
+                consistency_level: self
+                    .options
+                    .consistency_level
+                    .unwrap_or(self.consistency_level as i32),
                 use_default_consistency: self.options.consistency_level.is_none(),
                 expr_template_values: self.options.expr_template_values.clone(),
             })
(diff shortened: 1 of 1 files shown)


src/iterator.rs
```diff
@@ -554,7 +554,10 @@ impl QueryIterator {
                 guarantee_timestamp: 0,
                 query_params: init_ts_params,
                 not_return_all_meta: false,
-                consistency_level: self.options.consistency_level.unwrap_or(0),
+                consistency_level: self
```
liliu-z, line 557: When `options.consistency_level` is `None` and the collection default is `Customized`, this PR changes the wire value from `0` (Strong) to `4` (Customized) while `use_default_consistency` remains `true`. The iterator never consumes `options.guarantee_timestamp` (dead field at L72, L197), so the request carries `Customized` without the user-supplied timestamp that the proto spec requires (proto L1041). Whether this causes issues depends on server-side handling of `use_default_consistency=true` — which cannot be verified from this repo. The fix direction is correct for the 4 common levels (Strong/Session/Bounded/Eventually). Same pattern at L715, L961, L1403. Consider a follow-up to wire `options.guarantee_timestamp` into the request, or defensively reject `Customized` when the user hasn't set a consistency level.

yhmo:

PR #107, opened 2026-02-25
feat(query): make hybrid_search ranker optional with RRF default

marlon-costa-dc:
## Summary
- Makes the `ranker` parameter in `hybrid_search` an `Option<Box<dyn BaseRanker>>`.
- If `None` is provided, it falls back to `options.ranker` or defaults to `RrfRanker` with `k=60.0`.
- This simplifies the API for the common case where users just want standard RRF ranking.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -30,7 +30,7 @@
 //! let req2 = AnnSearchRequest::new(vec![vector2], "field2".to_string(), params2, 10);
 //! let ranker = WeightedRanker::new(vec![0.7, 0.3]);
 //!
-//! let results = client.hybrid_search("my_collection", vec![req1, req2], Box::new(ranker), None).await?;
+//! let results = client.hybrid_search("my_collection", vec![req1, req2], Some(Box::new(ranker)), None).await?;
 //! ```
 
 use std::collections::HashMap;
@@ -387,7 +387,7 @@ pub type HybridSearchOptions = SearchOptions;
 ///     .limit(100)
 ///     .offset(0);
 /// ```
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub struct QueryOptions {
     output_fields: Vec<String>,
     partition_names: Vec<String>,
@@ -412,20 +412,6 @@ pub enum IdType {
     Int64(Vec<i64>),
     /// String IDs (VarChar)
     VarChar(Vec<String>),
-}
-
-impl Default for QueryOptions {
-    fn default() -> Self {
-        Self {
-            output_fields: Vec::new(),
-            partition_names: Vec::new(),
-            guarantee_timestamp: 0,
-            query_params: vec![],
-            consistency_level: 0,
-            use_default_consistency: false,
-            expr_template_values: HashMap::new(),
-        }
-    }
 }
 
 impl QueryOptions {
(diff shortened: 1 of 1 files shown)


src/query.rs
```diff
@@ -1418,19 +1414,25 @@ impl Client {
     /// );
     ///
     /// let ranker = WeightedRanker::new(vec![0.7, 0.3]);
-    /// let results = client.hybrid_search("my_collection", vec![req1, req2], Box::new(ranker), None).await?;
+    /// let results = client.hybrid_search("my_collection", vec![req1, req2], Some(Box::new(ranker)), None).await?;
```
liliu-z, line 1417: The new doc example at L1417 moves `req1` and `req2` into `vec![req1, req2]`, then L1420 reuses them in a second call. `AnnSearchRequest` derives `Clone` but the example never calls `.clone()`, so this demonstrates use-after-move to users. Either split into independent examples or clone the requests before the second call.

src/query.rs
```diff
@@ -1721,7 +1723,7 @@ impl Client {
             IdType::VarChar(ids_string) => ids_string,
         };
 
-        let ids: Vec<String> = ids.into_iter().map(|x| x.into()).collect();
+        let ids: Vec<String> = ids.into_iter().collect();
```
liliu-z, line 1726: PR simplified this line from `.map(|x| x.into()).collect()` to `.into_iter().collect()`, but `ids` is already `Vec<String>`, so the collect is still a pointless identity reallocation. Since the PR intentionally touched this line to simplify it, the simplification should be completed by deleting the line entirely.

src/query.rs
```diff
@@ -1418,19 +1414,25 @@ impl Client {
     /// );
     ///
     /// let ranker = WeightedRanker::new(vec![0.7, 0.3]);
-    /// let results = client.hybrid_search("my_collection", vec![req1, req2], Box::new(ranker), None).await?;
+    /// let results = client.hybrid_search("my_collection", vec![req1, req2], Some(Box::new(ranker)), None).await?;
+    /// // Or pass ranker via options; if omitted, RRF with k=60 is used
+    /// let options = SearchOptions::new().ranker(Box::new(WeightedRanker::new(vec![0.7, 0.3])));
+    /// let results = client.hybrid_search("my_collection", vec![req1, req2], None, Some(options)).await?;
     /// ```
     pub async fn hybrid_search<S>(
```
liliu-z, line 1422: The third parameter of `hybrid_search()` changed from `Box<dyn BaseRanker>` to `Option<Box<dyn BaseRanker>>`. All existing external callers must wrap their argument with `Some(...)` or their code will fail to compile. While `0.1.0` SemVer allows breaking changes and the error is caught at compile time, this should be explicitly declared in release notes so downstream users are aware.

src/query.rs
```diff
@@ -1418,19 +1414,25 @@ impl Client {
...
         collection_name: S,
         reqs: Vec<AnnSearchRequest>,
-        ranker: Box<dyn BaseRanker>,
+        ranker: Option<Box<dyn BaseRanker>>,
         options: Option<HybridSearchOptions>,
     ) -> Result<Vec<SearchResult<'_>>>
     where
         S: Into<String>,
     {
         let options = options.unwrap_or_default();
```
liliu-z, line 1432: The new priority logic `ranker.or(options.ranker).unwrap_or_else(|| RrfRanker(60.0))` at L1432-1435 introduces three distinct branches (explicit parameter, options.ranker, default RRF) but none are covered by unit or integration tests. This core public API behavior is vulnerable to accidental regression during future refactoring.

yhmo:

Issue #113, opened 2026-04-01
[Feature]: Support Milvus 2.6.x API


richzw:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Currently, this rust sdk only support milvus 2.5. The new milvus 2.6 should be supported.

yhmo:
ClientV2 already supports it: https://github.com/milvus-io/milvus-sdk-rust/pull/129

PR #115, opened 2026-04-16
fix mutation status check

wg1026688210:
fix mutation status check
issue: https://github.com/milvus-io/milvus-sdk-rust/issues/113


wg1026688210:
@yah01  cc

PR #123, opened 2026-06-08
Support list_persistent_segments/list_loaded_segments

yhmo:


tests/utility.rs
```diff
@@ -55,6 +55,31 @@ async fn flush_all_and_get_state() -> Result<()> {
...
 }
 
+#[tokio::test]
+async fn list_persistent_segments() -> Result<()> {
+    let (client, collection) = create_test_collection(true).await?;
+    let collection_name = collection.name().to_string();
+
+    run_with_collection_cleanup(&client, vec![collection_name], || async {
+        client.flush(collection.name()).await?;
+        client.list_persistent_segments(collection.name()).await?;
```
yhmo: The test currently only checks that the RPC succeeds. Since `create_test_collection` inserts and flushes data, can we assert that at least one persistent segment is returned? That would catch regressions where the wrapper calls the right endpoint but accidentally drops `resp.infos` or returns an empty result.

```suggestion
        let segments = client.list_persistent_segments(collection.name()).await?;
        assert!(!segments.is_empty());
```

tests/utility.rs
```diff
@@ -55,6 +55,31 @@ async fn flush_all_and_get_state() -> Result<()> {
...
+    .await
+}
+
+#[tokio::test]
+async fn list_loaded_segments() -> Result<()> {
+    let (client, collection) = create_test_collection(true).await?;
+    let collection_name = collection.name().to_string();
+
+    run_with_collection_cleanup(&client, vec![collection_name], || async {
+        client.list_loaded_segments(collection.name()).await?;
```
yhmo: Same here: the helper loads a collection with data, so the test can validate the new API returns segment info instead of only checking for a successful status. This would make the regression coverage stronger.

```suggestion
        let segments = client.list_loaded_segments(collection.name()).await?;
        assert!(!segments.is_empty());
```

yhmo (commented):
Non-blocking test coverage suggestion for the new segment listing APIs.

yhmo (commented):
I found one issue to address before merging.

src/collection.rs
```diff
@@ -414,6 +414,32 @@ impl Client {
...
+    where
+        S: Into<String>,
+        I: IntoIterator<Item = &'a String>,
+    {
+        let partition_names: Vec<String> = partition_names.into_iter().map(|x| x.into()).collect();
+        let resp = self
+            .client
+            .clone()
+            .get_loading_progress(proto::milvus::GetLoadingProgressRequest {
+                base: Some(MsgBase::new(MsgType::LoadCollection)),
```
yhmo: This shared helper now hard-codes MsgType::LoadCollection, but load_partitions() also calls it. That changes GetLoadingProgressRequest.base.msg_type from the previous LoadPartitions value to LoadCollection for partition loads. If Milvus validates or routes on base.msg_type, partition progress checks can now be misclassified or rejected. Please preserve the caller-specific msg type (for example by passing MsgType into the helper, or by keeping separate helpers for collection vs partition loading).

PR #124, opened 2026-06-09
Auto-start/stop milvus container for test cases

yhmo:


yhmo (commented):
I found two issues to address before merging.

tests/common/mod.rs
```diff
@@ -147,6 +183,206 @@ pub async fn create_test_collection_custom(
...
+            .retain(|tracked_pid| process_is_alive(*tracked_pid));
+
+        if state.pids.is_empty() {
+            if !milvus_healthy() {
+                docker_compose_up(&compose_file)?;
+                wait_for_milvus()?;
+                state.managed = true;
+            } else {
+                state.managed = false;
+            }
```
yhmo: This cleanup registration only exists under `#[cfg(unix)]`, so on non-Unix platforms the bootstrap starts Docker but never tears it down. PR 124 also removes the workflow-level `docker compose down` management, so this leaks containers for any future Windows/local runs that hit the bootstrap. Please either add an explicit non-Unix cleanup path (for example via a Windows atexit registration or a scoped guard object with `Drop`) or keep cleanup external on platforms where you do not register it.

yhmo (commented):
I found two issues to address before merging.

scripts/run_tests.sh
```diff
@@ -0,0 +1,39 @@
...
+
+  echo "Timed out waiting for Milvus health endpoint: $HEALTH_URL" >&2
+  return 1
+}
+
+trap cleanup EXIT INT TERM
+
+docker compose -f "$COMPOSE_FILE" up -d
+wait_for_milvus
+
```
yhmo, line 32: When `CODE_COV=true`, this runs `cargo llvm-cov` instead of plain `cargo test`, which is what CI wants. But in the default branch it still always executes `cargo test "$@"`. That means a call like `./scripts/run_tests.sh env RUST_BACKTRACE=1 cargo test --test client --no-run` will be turned into `cargo test env RUST_BACKTRACE=1 cargo test --test client --no-run`, which is not the intended wrapper behavior. Please either keep the earlier “run exactly `$@` when args are provided” behavior, or add an explicit flag/argument parser so the script can safely support both `cargo test` and arbitrary wrapped commands.

tests/testsoutsideoftestsuite/client_with_timeout.rs
```diff
@@ -1,6 +1,8 @@
 use milvus::client::*;
 use std::time::Duration;
 
+#[path = "../common/mod.rs"]
```
yhmo: This test now includes `../common/mod.rs`, but it still constructs `Client::with_timeout(...)` directly and never uses anything from `common`. As written this import does not help the Docker lifecycle and only adds an extra compiled module plus warnings. If the wrapper script is now the intended Docker orchestration mechanism, please remove the unused import/module. If in-test bootstrap is still intended, the test needs to actually call the helper before constructing the client.

PR #125, opened 2026-06-11
Add missed index types and metric types, remove deprecated metric types

yhmo:


yhmo (commented):
Found 1 correctness issue in enum parsing that can mis-describe server-reported metric types.

src/index/mod.rs
```diff
@@ -71,10 +75,13 @@ pub enum MetricType {
     COSINE,
     HAMMING,
     JACCARD,
-    TANIMOTO,
-    SUBSTRUCTURE,
-    SUPERSTRUCTURE,
+    MHJACCARD,
     BM25,
+    MAX_SIM_COSINE,
```
yhmo, line 85: IndexInfo::from currently parses metric_type with MetricType::from_str(...) and falls back to MetricType::L2 on failure. If Milvus returns MAX_SIM for struct indexes, this change will mis-describe the index as L2. Please add MAX_SIM as an alias of MAX_SIM_COSINE and cover parsing MAX_SIM in the unit test.

yhmo (commented):
Found 1 remaining correctness issue around parsing legacy server-reported index types.

src/index/mod.rs
```diff
@@ -22,26 +22,14 @@ pub enum IndexType {
...
-    RHNSWSQ,
-    #[strum(serialize = "IVF_HNSW")]
-    IvfHNSW,
-    #[strum(serialize = "ANNOY")]
-    ANNOY,
-    #[strum(serialize = "NGT_PANNG")]
-    NGTPANNG,
-    #[strum(serialize = "NGT_ONNG")]
-    NGTONNG,
+    #[strum(serialize = "HNSW_SQ")]
```
yhmo, line 29: Removing the old enum variants here changes `describe_index` behavior for existing indexes. `IndexInfo::from` still does `IndexType::from_str(...).unwrap_or(IndexType::Flat)`, so any server-reported value like `IVF_SQ8_HYBRID`, `NSG`, `RHNSW_*`, `ANNOY`, or `NGT_*` will now be silently rewritten to `Flat`. Even if we do not want these in new index creation, we should keep parsing support for server-returned values (or surface an error instead of defaulting to `Flat`).

PR #126, opened 2026-06-11
Support CDC interfaces

yhmo:


tests/cdc.rs
```diff
@@ -0,0 +1,34 @@
+use milvus::client::Client;
```
yhmo, line 2: Middle ---- `tests/cdc.rs:1`

This new test file is not actually part of the project test suite. This repo enumerates integration tests explicitly in `Cargo.toml` via `[[test]]` entries, and there is no entry for `tests/cdc.rs`, so these CDC tests will never run in CI or under a normal `cargo test` invocation. Please either add a `[[test]]` target for `tests/cdc.rs` or move these checks into an already-registered integration test target.

src/cdc.rs
```diff
@@ -0,0 +1,80 @@
...
+use crate::proto::common::{ReplicateCheckpoint, ReplicateConfiguration};
+use crate::proto::milvus::{
+    GetReplicateConfigurationRequest, GetReplicateInfoRequest,
+    UpdateReplicateConfigurationRequest,
+};
+use crate::error::Result;
+use crate::utils::status_to_result;
+
+impl Client {
+    /// Retrieves replication metadata for a target pchannel.
```
yhmo: High ---- `src/cdc.rs:29`

`GetReplicateInfoResponse` includes both `checkpoint` and `salvage_checkpoint`, but this Rust API collapses the whole response down to a single `ReplicateCheckpoint`. That drops the salvage checkpoint entirely, which is specifically needed after force-promote/data-salvage flows. Java preserves both fields in `GetReplicateInfoResp`, so the Rust surface is currently losing part of the server response. Please return a response type that exposes both checkpoints instead of only `resp.checkpoint`.

src/cdc.rs
```diff
@@ -0,0 +1,239 @@
...
+    }
+}
+
+impl Client {
+    /// Retrieves replication checkpoint metadata for a target pchannel.
+    pub async fn get_replicate_info<S1: Into<String>, S2: Into<String>>(
+        &self,
+        source_cluster_id: S1,
+        target_pchannel: S2,
+    ) -> Result<(Option<ReplicateCheckpoint>, Option<ReplicateCheckpoint>)> {
```
yhmo: High ---- `GetReplicateInfoResponse` in `milvus.proto` currently only carries a single `checkpoint`, but this public API already returns a 2-tuple and hard-codes the second slot to `None`. That means the SDK is exposing impossible state to callers: they have to handle a second checkpoint that the wire format can never populate. Please either reduce this return type to a single `Option<ReplicateCheckpoint>` for now, or introduce a named response type only when the server/proto actually exposes a second field.

tests/cdc.rs
```diff
@@ -0,0 +1,36 @@
+use milvus::cdc::ReplicateConfiguration;
+use milvus::client::Client;
+use milvus::error::Result;
+
+mod common;
+use common::*;
+
+#[tokio::test]
+#[ignore = "replication configuration APIs require a replication-enabled Milvus deployment"]
```
yhmo, line 9: Medium ---- All of the new CDC tests in this file are `#[ignore]`, so the new CDC code paths added in `src/cdc.rs` still get no runnable coverage in CI. Since most of that module is pure conversion logic, it would be worth adding a few non-ignored unit tests for the `From` mappings and fallback branches (for example unknown `wal_name` and missing `connection_param`) so this API surface is exercised without needing a replication-enabled Milvus deployment.

PR #127, opened 2026-06-16
Reorg source code as v1

yhmo:


src/v1/collection.rs
```diff
@@ -72,22 +73,107 @@ pub struct DescribeCollection {
...
+static GLOBAL_TIMESTAMPS: OnceLock<Arc<dashmap::DashMap<String, Timestamp>>> = OnceLock::new();
+
+fn global_timestamps() -> Arc<dashmap::DashMap<String, Timestamp>> {
+    Arc::clone(GLOBAL_TIMESTAMPS.get_or_init(|| Arc::new(dashmap::DashMap::new())))
+}
+
+static DEFAULT_COLLECTION_CONSISTENCY: OnceLock<ConsistencyLevel> = OnceLock::new();
+
+fn default_collection_consistency(collection: &Collection) -> ConsistencyLevel {
+    *DEFAULT_COLLECTION_CONSISTENCY.get_or_init(|| collection.consistency_level)
```
yhmo: src/v1/collection.rs line:98
High ---- `DEFAULT_COLLECTION_CONSISTENCY` memoizes the first collection\x27s consistency level process-wide. After the first `describe_collection`, every later query/search that relies on the collection default will reuse that same level, so a `Session` collection can silently behave like `Strong` (or vice versa). This needs to read `collection.consistency_level` per request instead of caching a single global value.

src/v1/collection.rs
```diff
@@ -72,22 +73,107 @@ pub struct DescribeCollection {
...
 impl CollectionCache {
     pub fn new(
         client: MilvusServiceClient<InterceptedService<Channel, CombinedInterceptor>>,
     ) -> Self {
         Self {
-            collections: dashmap::DashMap::new(),
-            timestamps: dashmap::DashMap::new(),
-            client: client,
+            collections: Arc::new(dashmap::DashMap::new()),
+            timestamps: global_timestamps(),
```
yhmo: src/v1/collection.rs line:152
High ---- `CollectionCache::new()` now points every `Client` at a process-global timestamp map. That leaks session-consistency state across unrelated clients, and operations that clear/remove timestamps on one client (`using_database`, `drop_collection`, `truncate_collection`) can erase another client\x27s last-write timestamp, causing later `Session` reads to fall back to `1` instead of the caller\x27s latest write timestamp.

src/v1/query.rs
```diff
@@ -613,21 +611,13 @@ impl QueryOptions {
...
-    ///
-    /// * `use_default_consistency` - Boolean flag
-    ///
-    /// # Returns
-    ///
-    /// Self for method chaining
-    pub fn use_default_consistency(mut self, use_default_consistency: bool) -> Self {
-        self.use_default_consistency = use_default_consistency;
+    /// Clears the explicit consistency level so the server uses the collection default.
+    pub fn use_default_consistency(mut self) -> Self {
```
yhmo: src/v1/query.rs line:619
Medium ---- Changing `use_default_consistency` from `fn(bool)` to a zero-arg method is a source-breaking API change for existing SDK users. Any caller that currently does `use_default_consistency(true/false)` will stop compiling even though this PR is otherwise framed as a reorg, so this should keep the old signature or provide a compatibility path.

src/v1/query.rs
```diff
@@ -1592,21 +1603,18 @@ impl Client {
...
-                .unwrap_or(ConsistencyLevel::default() as _),
-            use_default_consistency: extract_param(
-                &options.search_params,
-                "use_default_consistency",
-                "true",
-            )
-            .parse()
-            .unwrap_or(true),
+            consistency_level: options.consistency_level.unwrap_or_default(),
+            use_default_consistency: options.consistency_level.is_none(),
```
liliu-z: Hybrid search derives use_default_consistency solely from consistency_level — outer request `use_default_consistency: options.consistency_level.is_none()` (line 1617) and each sub-request `use_default_consistency: !search_params.iter().any(|p| p.key == "consistency_level")` (line 1581) — so a raw `add_param("use_default_consistency", "false")` is silently ignored where the old extract_param path honored it. Narrow (the default stays true; only an explicit opt-out without a consistency_level differs) but silent, and inconsistent with add_param's explicit backward-compat shim for consistency_level.

src/v1/database.rs
```diff
@@ -531,7 +531,11 @@ impl Client {
         let new_client =
             MilvusServiceClient::with_interceptor(self.channel.clone(), combined_interceptor);
         self.client = new_client.clone();
-        self.collection_cache = CollectionCache::new(new_client);
+        self.collection_cache = CollectionCache::with_shared_state(
```
liliu-z, line 534: using_database() rebuilds the cache via `CollectionCache::with_shared_state(...)` (line 534) to preserve collections/timestamps, but the preceding `self.collection_cache.clear()` (line 517) empties both maps first, so the new plumbing only ever carries empty Arcs. Separately, collections are keyed by bare name (collection.rs:229 `self.collections.remove(collection_name)`) while timestamps use the db-scoped `cache_key(db_name, name)` (collection.rs:222) — an inconsistent scheme that becomes a cross-db collision hazard if that clear() is ever removed. Not a runtime bug today; align the keying and reconsider clearing under the new shared-state design.

src/v1/query.rs
```diff
@@ -613,21 +611,15 @@ impl QueryOptions {
...
-    ///
-    /// * `use_default_consistency` - Boolean flag
-    ///
-    /// # Returns
-    ///
-    /// Self for method chaining
-    pub fn use_default_consistency(mut self, use_default_consistency: bool) -> Self {
-        self.use_default_consistency = use_default_consistency;
+    /// Controls whether the server uses the collection default consistency.
+    pub fn use_default_consistency(mut self, enabled: bool) -> Self {
```
yhmo, line 607: src/v1/query.rs line:619
Medium ---- `use_default_consistency(false)` no longer has any effect here. This method only clears `consistency_level` when `enabled` is true, and `query()` later derives the request flag from `options.consistency_level.is_none()`. Callers that used this API to opt out of a collection's default consistency will now still send `use_default_consistency=true`, which silently changes query visibility on Session collections.

src/v1/query.rs
```diff
@@ -1357,12 +1374,18 @@ impl Client {
...
+                    0,
+                    options.consistency_level,
+                    options.consistency_level.is_none(),
+                )
+                .await,
                 not_return_all_meta: false,
-                consistency_level: ConsistencyLevel::default() as _,
-                use_default_consistency: false,
+                consistency_level: options.consistency_level.unwrap_or_default(),
+                use_default_consistency: options.consistency_level.is_none(),
```
yhmo: src/v1/query.rs line:1388
Medium ---- `search()` now ignores a raw `search_params` opt-out for `use_default_consistency`. The previous code extracted this flag from `search_params`, and `hybrid_search()` in this PR still does via `request_uses_default_consistency(...)`, but this path hardcodes the request field from `options.consistency_level.is_none()`. Existing callers that do `SearchOptions::add_param("use_default_consistency", "false")` will silently switch back to collection-default consistency here.

src/v1/query.rs
```diff
@@ -1161,13 +1172,14 @@ impl Client {
         collection_name: &str,
         consistency_level: ConsistencyLevel,
     ) -> u64 {
+        let db_name = self.db_name.as_deref().unwrap_or("default");
```
liliu-z: The consistency refactor replaced every caller of `get_gts_from_consistency` with `CollectionCache::guarantee_timestamp_for_request`, yet this PR still modified the function (adding the `db_name` lookup at line 1175) and left it in place. It now has zero callers anywhere in the crate, so it is dead `pub(crate)` code that will emit an unused-function warning. Remove it as part of the reorg.

src/v1/query.rs
```diff
@@ -1224,20 +1194,19 @@ impl Client {
...
                 query_params: options.query_params.clone(),
                 not_return_all_meta: false,
-                consistency_level: if options.consistency_level > 0 {
-                    options.consistency_level
-                } else {
-                    0
-                },
-                use_default_consistency: options.use_default_consistency,
+                consistency_level: options.consistency_level.unwrap_or_default(),
+                use_default_consistency: options.consistency_level.is_none(),
```
yhmo, line 1209: src/v1/query.rs line:1209
Medium ---- `use_default_consistency(false)` no longer has any effect here. This request now always derives `use_default_consistency` from `options.consistency_level.is_none()`, so callers that explicitly opt out of collection-default consistency still send `true` unless they also set a dedicated consistency level. That silently changes query visibility semantics on Session collections and is a source-breaking behavior change for existing SDK users.

PR #128, opened 2026-06-29
Support get_compaction_plans/alter_role/update_user

yhmo:


src/v1/authentication.rs
```diff
@@ -258,6 +258,7 @@ impl Client {
                 base: Some(MsgBase::new(MsgType::CreateRole)),
                 entity: Some(proto::milvus::RoleEntity {
                     name: role_name.into(),
+                    description: String::new(),
```
yhmo, line 262: src/v1/authentication.rs line:261
High ---- The pinned milvus-proto revision on this branch still defines RoleEntity with only a name field, so every added description initializer here (and the same pattern in describe_role/grant/revoke privilege) fails to compile against the generated proto types on this PR branch.

src/v1/authentication.rs
```diff
@@ -1016,6 +1022,33 @@ impl Client {
...
+            .client
+            .clone()
+            .update_credential(proto::milvus::UpdateCredentialRequest {
+                base: Some(MsgBase::new(MsgType::UpdateCredential)),
+                username: user_name.into(),
+                old_password: String::new(),
+                new_password: String::new(),
+                created_utc_timestamps: 0,
+                modified_utc_timestamps: 0,
+                description: Some(description.into()),
```
yhmo, line 1045: src/v1/authentication.rs line:1044
High ---- UpdateCredentialRequest in the pinned milvus-proto revision does not have a description field, so this new update_user path—and the matching description: None added in update_password below—won't build on the target branch.

src/v1/authentication.rs
```diff
@@ -1066,6 +1099,30 @@ impl Client {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `role_name` - The role to update
+    /// * `description` - The new role description
+    pub async fn alter_role<S: Into<String>>(&self, role_name: S, description: S) -> Result<()> {
+        let res = self
+            .client
+            .clone()
+            .alter_role(proto::milvus::AlterRoleRequest {
```
yhmo, line 1123: src/v1/authentication.rs line:1122
High ---- The pinned milvus-proto submodule for this branch does not define an AlterRoleRequest message or AlterRole RPC, so this new method cannot compile until the proto dependency is updated in the same PR.

src/v1/authentication.rs
```diff
@@ -1016,6 +1023,33 @@ impl Client {
...
+            .client
+            .clone()
+            .update_credential(proto::milvus::UpdateCredentialRequest {
+                base: Some(MsgBase::new(MsgType::UpdateCredential)),
+                username: user_name.into(),
+                old_password: String::new(),
+                new_password: String::new(),
+                created_utc_timestamps: 0,
+                modified_utc_timestamps: 0,
+                description: Some(description.into()),
```
yhmo, line 1045: src/v1/authentication.rs line:1045
Medium ---- `update_user` and `alter_role` write descriptions, but the SDK still has no corresponding read surface: `describe_user` drops `UserResult.description`, and `describe_role` only rebuilds privilege data from `select_grant`. That leaves callers unable to observe the metadata they just set through the Rust API.

Issue #64, opened 2023-10-16
[Feature]: Support Count API


jiaoew1991:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Milvus support `Count` API from 2.3.0, we need to support it,
more details about the it can be found here https://milvus.io/docs/query.md#Count-entities

Hardik-7892:
Hi, I'd like to pick this up — I've already implemented and tested it locally, so I'd appreciate being assigned to avoid duplicate work.

Summary of what's done:
- Client::count() (whole-collection) and Client::count_with_expr() (filtered, honoring QueryOptions like partition names/consistency) in src/v1/query.rs, using the existing count(*) Query aggregate — no proto bump needed since the pinned proto has no dedicated Count RPC.
- Unknown/empty results degrade to 0 instead of panicking.
- Integration test count_api added in tests/v1/milvus26.rs, passing against a live Milvus v2.6.19.
- Commit is DCO-signed.

I'll open the PR from my fork shortly and drop the link here. Thanks!

xiaofan-luan:
I think we don't need a count api?

it should be already part of aggregation.

Hardik-7892:
Agreed, count(\*) is already reachable via the Query aggregation, and that's exactly what PR #133 does under the hood: count()/count_with_expr() are thin wrappers around the existing aggregate, matching pymilvus.num_entities and milvus-sdk-go Count. No new RPC or proto change. If the team prefers, I'm happy to drop the public API and keep just the documented count(\*) Query pattern instead.

yhmo:
ClientV2 already supports it: https://github.com/milvus-io/milvus-sdk-rust/pull/129

PR #129, opened 2026-07-15
Rust SDK V2

yhmo:
- Add src/v2/ for classes of V2. A struct ClientV2 for V2 interfaces.
- Add src/v2/request/ for requests of V2, categories by function areas.
- Add src/v2/response/ for responses of V2, categories by function areas.
- Don't reuse any structs/enums/methods in src/v1/
- Add tests/v2/ut/ for V2 unittest cases, add tests/v2/st/ for V2 end-to-end test cases
- Add examples/v2/ for examples of V2
- Ensure V1 is not impacted so that user's old client code still work

PR #133, opened 2026-08-05
feat: support Count API (#64)

Hardik-7892:
Related issue: #64

Adds whole-collection and filtered entity counting to the Rust SDK,
matching pymilvus/milvus-sdk-go:

- `Client::count(collection_name)` — whole-collection count
- `Client::count_with_expr(collection_name, expr, &QueryOptions)` —
  filtered count honoring options (partition_names, consistency_level,
  guarantee_timestamp, namespace)

Implemented via the existing `count(*)` Query aggregate — the pinned
proto has no dedicated `Count` RPC, so no proto bump is needed.
Unknown/empty results degrade to `0` instead of panicking.

## Test
- `cargo check --lib` / `cargo check --tests` pass
- Integration test `count_api` in `tests/v1/milvus26.rs`, passing
  against a live Milvus **v2.6.19** (whole count == 300, filtered count
  == 300)

Signed-off-by already on the commit (DCO).

### DIFF
--- a/src/v1/query.rs
+++ b/src/v1/query.rs
@@ -49,7 +49,7 @@ use crate::proto::milvus::{QueryRequest, SearchRequest};
 use crate::proto::schema::DataType;
 use crate::v1::types::Field;
 use crate::utils::status_to_result;
-use crate::v1::value::Value;
+use crate::v1::value::{Value, ValueVec};
 use crate::{error::*, proto};
 
 /// Represents an ANN (Approximate Nearest Neighbor) search request
@@ -358,6 +358,25 @@ impl BaseRanker for RrfRanker {
 ///
 /// Type alias for SearchOptions used in hybrid search operations
 pub type HybridSearchOptions = SearchOptions;
+
+/// Special output field that makes the server compute an entity count
+/// aggregate instead of returning rows.
+const COUNT_FIELD: &str = "count(*)";
+
+/// Extracts the `count(*)` aggregate value from a query result.
+///
+/// The server returns the count as a single integer-backed column, so the value
+/// is read as the first element of a `ValueVec::Long`. Any unexpected shape or
+/// empty result degrades to `0` instead of panicking.
+fn parse_count(results: &[FieldColumn]) -> i64 {
+    results
+        .first()
+        .and_then(|col| match &col.value {
+            ValueVec::Long(values) => values.first().copied(),
+            _ => None,
+        })
+        .unwrap_or(0)
+}
 
 /// QueryOptions for client.query()
 ///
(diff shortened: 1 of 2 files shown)


src/v1/query.rs
```diff
@@ -1218,6 +1237,72 @@ impl Client {
...
+    /// # Returns
+    ///
+    /// The number of entities in the collection.
+    ///
+    /// # Example
+    ///
+    /// ```rust,ignore
+    /// let count = client.count("my_collection").await?;
+    /// ```
+    pub async fn count<S>(&self, collection_name: S) -> Result<i64>
```
yhmo, line 1272: src/v1/query.rs line:1258
Medium ---- This adds a new public API to the deprecated V1 compatibility surface even though counting is already supported through Query aggregation (and the linked maintainer feedback explicitly questions a separate Count API). Once published, these methods become another compatibility contract that must track Query/server semantics—as the pagination mismatch below already demonstrates—so please keep this as the existing count(*) Query pattern or add the feature in the active V2 surface instead.

src/v1/query.rs
```diff
@@ -1218,6 +1237,72 @@ impl Client {
...
+        expr: &str,
+        options: &QueryOptions,
+    ) -> Result<i64>
+    where
+        S: Into<String>,
+    {
+        let mut opts = options.clone();
+        opts.output_fields = vec![COUNT_FIELD.to_string()];
+        let results = self.query(collection_name, expr, &opts).await?;
+        Ok(parse_count(&results))
```
yhmo: src/v1/query.rs line:1303
Medium ---- Returning Ok(0) here makes a missing field, wrong protobuf type, or empty aggregate vector indistinguishable from a genuinely empty collection. Client::query has already established that the RPC/status succeeded, so these shapes are malformed response data and should return Error::Unexpected/Conversion (and ideally validate the count(*) column) instead of silently reporting a false count that callers may act on.

src/v1/query.rs
```diff
@@ -1218,6 +1237,72 @@ impl Client {
...
+    pub async fn count_with_expr<S>(
+        &self,
+        collection_name: S,
+        expr: &str,
+        options: &QueryOptions,
+    ) -> Result<i64>
+    where
+        S: Into<String>,
+    {
+        let mut opts = options.clone();
```
yhmo, line 1318: src/v1/query.rs line:1300
Medium ---- Cloning the full QueryOptions also forwards query_params such as limit. Milvus 2.6 explicitly rejects count(*) with pagination ("count entities with pagination is not allowed"), so a type-correct call like count_with_expr(..., &QueryOptions::new().limit(10)) fails even though this method accepts general QueryOptions. Please reject incompatible fields locally, filter pagination/iterator parameters, or expose a count-specific options type so every accepted configuration forms a valid request.

tests/v1/milvus26.rs
```diff
@@ -179,6 +179,43 @@ async fn truncate_collection() -> Result<()> {
...
+
+        // Actual entity count should equal the number of inserted rows
+        let whole = client.count(&collection_name).await?;
+        assert_eq!(whole, entity_count, "whole-collection count");
+
+        // count_with_expr with an always-true filter matches every row
+        let filtered = client
+            .count_with_expr(
+                &collection_name,
+                "id >= 0 || id < 0",
```
yhmo: tests/v1/milvus26.rs line:208
Low ---- This predicate is true for every i64, so the test still passes if count_with_expr accidentally drops expr and performs an unfiltered count. Please use deterministic IDs and a filter with a known strict subset (and assert that subset) so the new filtered-count behavior is actually protected.

Hardik-7892:
Hi @yhmo , thanks for the detailed review. All 4 points are addressed in the follow-up commit:

1. parse_count now returns Result<i64> and raises Error::Unexpected for a missing count(*) column, a non-integer column, or an empty aggregate — no more silent 0. I verified against a live empty loaded collection that a genuine count of 0 still comes back as a 0 row, so the empty-collection case is unaffected.

2. count_with_expr now strips pagination/iterator query_params (limit, offset, iterator, and any iter_*) before querying, so a count(*) request can no longer be rejected for "count entities with pagination is not allowed". Documented on the method.

3. count_api now inserts deterministic sequential IDs 0..300 and asserts count() == 300 and count_with_expr("id < 150") == 150, so the filter is genuinely exercised — a dropped expr would fail the assertion.

4. V1 vs V2: Here I'd push back gently. Issue #64 was filed asking for exactly this and is tagged good first issue; pymilvus (Collection.num_entities) and milvus-sdk-go (Count) both expose a first-class count API layered on the same Query aggregation, and the Rust SDK has no equivalent today. That said, I can see #129 is still WIP and src/v2/ isn't on main yet, so there's no V2 surface to land this in right now — and since V2 is deliberately decoupled from V1, this addition doesn't conflict with it. When #129 merges I'm happy to port count() to the V2 ClientV2 as a follow-up; if you'd still rather not grow the V1 API at all, I'm happy to close this and we can add it to V2 later.


Tests pass against a live Milvus v2.6.19.

src/v1/query.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v1/query.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

tests/v1/milvus26.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v1/query.rs
yhmo: no need

PR #134, opened 2026-08-10
Fix CI problems

yhmo:
- Replaces Docker Compose with a Python-managed standalone Milvus container using embedded
    etcd, local storage, and Rocksmq.

  - Runs non-server tests before launching Milvus.
  - Fixes coverage generation and signal-safe container cleanup.
  - Adds Linux Docker setup and macOS 15 non-server testing.
  - Updates Mergify CI-label rules and development documentation.


scripts/install_deps.sh
```diff
@@ -99,15 +128,17 @@ install_rust() {
 
   if ! cargo llvm-cov --version >/dev/null 2>&1; then
     log "Installing cargo-llvm-cov"
-    cargo install cargo-llvm-cov --locked
+    if ! cargo install cargo-llvm-cov --locked; then
```
liliu-z, line 176: `cargo install cargo-llvm-cov --locked` now only emits a warning when it fails instead of aborting, and `run_tests.sh` has no preflight check for the binary. That removes fail-fast from Linux CI: the missing tool is no longer surfaced at dependency-install time, and the invariant that a successful install script implies a usable coverage toolchain no longer holds. The failure instead resurfaces much later — after Milvus has been booted — as `no such command: llvm-cov`, wasting a full CI cycle and pointing at the wrong stage. Either restore the hard failure or add a preflight `command -v cargo-llvm-cov` guard in `run_tests.sh`.

scripts/install_deps.sh
```diff
@@ -116,11 +147,27 @@ verify_required_tools() {
     fi
   done
 
-  if ! cargo llvm-cov --version >/dev/null 2>&1; then
-    echo "Missing cargo-llvm-cov" >&2
+  if ! command -v pkg-config >/dev/null 2>&1 && ! command -v pkgconf >/dev/null 2>&1; then
```
liliu-z: The new `exit 1` gate requires `pkg-config`/`pkgconf`, but the resolved dependency graph never invokes it — there is no `openssl-sys`, no `native-tls`, and no `pkg-config` crate, only the pure-Rust `openssl-probe`. Gating on a tool that is not actually a build input turns a non-requirement into a hard blocker. On macOS where Homebrew has `pkgconf` installed but unlinked, the check fails and the script aborts on an otherwise perfectly valid setup, blocking developers from installing dependencies at all. Drop the gate or downgrade it to a warning.

scripts/install_deps.sh
```diff
@@ -117,10 +145,17 @@ verify_required_tools() {
   done
 
   if ! cargo llvm-cov --version >/dev/null 2>&1; then
-    echo "Missing cargo-llvm-cov" >&2
-    missing=1
+    warn "Missing cargo-llvm-cov; CODE_COV=true coverage reports will be unavailable"
   fi
 
+  for tool in lcov genhtml; do
```
yhmo: scripts/install_deps.sh line:151
Low ---- The current verifier still treats the `lcov` executable as a coverage prerequisite, but `run_tests.sh` only invokes `cargo llvm-cov --lcov` (which uses `llvm-cov export`) and `genhtml`. A setup with those two commands but no `lcov` will work while this loop warns that coverage is unavailable. Please drop the `lcov` executable check or change the message so verification matches the actual consumer.

scripts/install_deps.sh
```diff
@@ -65,8 +69,27 @@ install_macos_packages() {
     exit 1
   fi
 
-  log "Installing build and coverage packages with Homebrew"
-  brew install lcov openssl@3 pkg-config
+  log "Checking build packages with Homebrew"
```
yhmo, line 118: scripts/install_deps.sh line:72
Low ---- Could we add direct coverage for this macOS installer branch? The current macOS workflow only runs `cargo test --no-run`, so it never executes the Homebrew logic this PR is fixing; a Bash/Homebrew regression here would pass every current check. A mocked-PATH shell test or an intentional installer smoke step would close that gap.

scripts/run_tests.sh
```diff
@@ -18,57 +18,133 @@
...
+  cargo test --test v2_ut -- --test-threads=4
+  cargo test --doc
+}
+
+run_non_server_coverage_tests() {
+  echo "==> Running compile checks and non-server coverage tests"
+  cd "$ROOT_DIR"
+  cargo check --all-targets
+  cargo llvm-cov clean --workspace
+  cargo llvm-cov --workspace --no-clean --no-report --lib -- --test-threads=4
```
yhmo: scripts/run_tests.sh line:77
High ---- `cargo llvm-cov` rejects `--no-clean` together with `--no-report`; the exact-head Ubuntu job already fails here with `error: --no-report may not be used together with --no-clean`, so neither coverage tests nor the server suite/report runs. Please use the supported accumulation form (for example, `--no-report` for each run followed by `cargo llvm-cov report`) without this incompatible option pair.

scripts/run_tests.sh
```diff
@@ -18,57 +18,133 @@
...
+  command -v python3 >/dev/null 2>&1 || {
+    echo "python3 is required to launch the Milvus test container" >&2
+    exit 1
+  }
 }
 
-trap cleanup EXIT INT TERM
+start_milvus() {
+  echo "==> Starting Milvus for server-backed tests"
+  trap cleanup EXIT
```
yhmo: scripts/run_tests.sh line:96
Medium ---- Registering cleanup only for `EXIT` drops the base script's `INT`/`TERM` handling. Bash does not execute an EXIT trap on default SIGTERM termination, so cancellation or timeout after `docker run -d` leaves the Milvus container alive on ports 19530/9091; please trap `INT` and `TERM` while preserving the original exit status.

README.md
```diff
@@ -331,13 +331,28 @@ cargo test -- --test-threads=4
...
-healthy, runs the test suite, and shuts the containers down afterward:
+`scripts/run_tests.sh` uses two threads for system tests by default. Override this only when the
+Milvus test host has enough resources:
+
+```shell
+SYSTEM_TEST_THREADS=4 ./scripts/run_tests.sh
+```
+
+On Linux, the repository test script first runs the library, V2 unit, and documentation tests.
+After those pass, `tests/v2/st/milvus_container.py` starts one standalone Milvus container with
```
yhmo, line 347: README.md line:342
Low ---- This new single-container flow conflicts with the unchanged Development prerequisites above, which still require Docker Compose and say Docker must be installed separately. This PR deletes Compose and makes `install_deps.sh` install and verify Docker on Linux, so please update those earlier lines to describe the actual platform-specific prerequisites.

tests/v2/st/milvus_container.py
```diff
@@ -0,0 +1,240 @@
...
+    config: Path,
+) -> str:
+    """Start Milvus with embedded etcd, local storage, and the default WAL."""
+    docker("info", capture_output=True)
+    if not config.is_file():
+        raise RuntimeError(f"Milvus test configuration does not exist: {config}")
+    ensure_image(image)
+
+    name = f"milvus-sdk-rust-{uuid.uuid4().hex[:12]}"
+    result = docker(
```
yhmo: tests/v2/st/milvus_container.py line:154
Medium ---- `docker run` creates the container before calling `ContainerStart`, and without `--rm` the Docker CLI leaves that created container behind if start fails (for example, when ports 19530 or 9091 are already bound). This call then raises before stdout returns the ID, so neither this function nor `run_tests.sh` can remove it, regressing the previous Compose cleanup for partial startup. Please clean up by the already-known `name` on `docker run` failure, or split create/start so cleanup owns the ID before start.

liliu-z:
**milvus_container.py:45** The annotation `subprocess.CompletedProcess[str]` is evaluated at runtime, and `CompletedProcess` only became subscriptable in Python 3.9; on Python 3.8 or 3.6 this raises `TypeError: 'type' object is not subscriptable` at import time. Because the module is imported by every server-backed test, the entire server test suite dies at collection on Ubuntu 20.04 (py3.8) and RHEL 8 (py3.6) rather than failing a single case. Quote the annotation (`"subprocess.CompletedProcess[str]"`) or add `from __future__ import annotations`.

.github/mergify.yml
```diff
@@ -26,30 +26,34 @@ pull_request_rules:
...
       - files~=\.(?!md|png)
     actions:
       label:
         remove:
           - ci-passed
+
+  - name: Add ci-passed when no code changes
+    conditions:
+      - base=main
+      - "-files~=^(?!(.*\\.md$|.*\\.png$|\\.gitignore$|codecov\\.yml$|LICENSE$|OWNERS$|\\.github/mergify\\.yml$)).+"
```
liliu-z, line 53: The `ci-passed` rule requires success checks that only run for paths excluded by the path filters, so a PR whose diff is confined to `examples/**`, non-markdown files under `tutorial/**`, or `configs/**` triggers no qualifying workflow and never satisfies the condition. Such PRs sit indefinitely without the label and cannot be auto-merged, requiring a manual label or an empty commit to unblock. Add a fallback rule that applies `ci-passed` when only these paths changed.

.github/mergify.yml
```diff
@@ -26,30 +26,34 @@ pull_request_rules:
...
 
-  - name: Remove ci-passed when CI failed
+  - name: Remove ci-passed when checks are not green
     conditions:
       - base=main
       - label=ci-passed
       - or:
-        - "-status-success=UT on Ubuntu-22.04"
-        - "-status-success=Compile on macOS 15"
+        - "check-pending=UT on Ubuntu-22.04"
```
liliu-z: The remove-label conditions test only `check-pending` and `check-failure`, but a check that has not been reported at all (e.g. after a new push that has not yet registered the run) is in neither state, so the condition never fires. A PR that already carries `ci-passed` keeps the label across a force-push or a new commit whose checks haven't started, letting unvalidated code be auto-merged. Also match on `check-success` being absent (or use `-check-success=...`) so the label is stripped whenever the check isn't affirmatively green.

scripts/install_deps.sh
```diff
@@ -44,17 +48,52 @@ install_linux_packages() {
...
+  if command -v systemctl >/dev/null 2>&1; then
+    log "Starting the Docker service"
+    "${SUDO[@]}" systemctl start docker || true
+  elif command -v service >/dev/null 2>&1; then
+    log "Starting the Docker service"
+    "${SUDO[@]}" service docker start || true
+  fi
+
+  if ! docker info >/dev/null 2>&1; then
+    echo "Docker is installed, but the current user cannot access a running Docker daemon. Start Docker and ensure this user has permission to use it." >&2
```
liliu-z, line 107: The Docker availability check calls `exit` on failure, and it is placed ahead of the `install_rust` step, so on any machine without Docker the script terminates before installing the toolchain it is primarily there to provide. A new contributor without Docker runs the setup script, sees a Docker error, and is left with no `cargo`/`rustc` at all — the failure is silent about what was skipped. Move the Docker check after `install_rust`, or downgrade it to a warning that lets the script continue.

src/v1/database.rs
```diff
@@ -22,13 +22,13 @@
...
 //!
-//! ```rust,ignore
+//! ```rust,no_run
 //! use milvus::v1::client::Client;
 //! use milvus::v1::database::CreateDbOptions;
 //!
 //! #[tokio::main]
 //! async fn main() -> Result<(), Box<dyn std::error::Error>> {
-//!     let client = Client::new("localhost:19530").await?;
+//!     let mut client = Client::new("localhost:19530").await?;
```
liliu-z: The doc example calls `Client::new("localhost:19530")` with no `http://` prefix, but the client parses its argument as a URL and rejects a schemeless authority. Anyone copying this snippet gets a URL parse error on the first call rather than a connection, and the same mistake appears at `resource_group.rs:32`. Use `http://localhost:19530` in both places.

liliu-z:
**milvus_container.py:170** Cleanup runs only through the normal teardown path, so a hard kill (SIGKILL, CI runner cancellation) leaves the container alive still holding the fixed host ports 19530 and 9091, and the label added for reaping is never queried by any code. The next run then fails at startup with `Bind for 0.0.0.0:19530 failed: port is already allocated`, and the operator must remove the container by hand. Add a startup sweep that removes containers carrying the reap label — matching on the label only, so concurrent live containers from other runs are not killed.

tests/v2/st/milvus_container.py
```diff
@@ -0,0 +1,267 @@
...
+        raise
+
+    return container_id
+
+
+def main() -> int:
+    signal.signal(signal.SIGTERM, handle_termination)
+    parser = argparse.ArgumentParser(description=__doc__)
+    subparsers = parser.add_subparsers(dest="command", required=True)
+
```
yhmo, line 243: tests/v2/st/milvus_container.py line:230
Medium ---- This still requires Python 3.7 even though `install_deps.sh` supports the yum path and only installs generic `python3` (RHEL 8 provides 3.6): Python 3.6 rejects `add_subparsers(required=True)`, and the same launcher also uses the 3.7-only `capture_output=` and `text=` subprocess keywords, so every server-backed run fails before Docker starts. Please either enforce/document Python >=3.7 in dependency verification or use the 3.6-compatible argparse/subprocess forms.

README.md
```diff
@@ -288,9 +289,10 @@ On supported Linux distributions and macOS, install the build and coverage tools
 ./scripts/install_deps.sh
 ```
 
-The script installs the Rust toolchain, formatting and linting components, `cargo-llvm-cov`,
-and the LCOV tools, including `lcov` and `genhtml`. Docker with the Compose plugin must be
-installed separately for server-backed tests.
+The script installs the Rust toolchain, formatting and linting components, `cargo-llvm-cov`, and
```
yhmo: README.md line:292
Low ---- The earlier coverage-tool failure was made non-fatal with a `run_tests.sh` preflight, but this new text still promises that `install_deps.sh` installs `cargo-llvm-cov` and `lcov`; the script now warns and exits successfully when either install fails. Please say it “attempts to install” the optional coverage tools (or otherwise note the best-effort behavior) so a successful setup run is not documented as guaranteeing coverage support.

scripts/install_deps.sh
```diff
@@ -99,15 +160,17 @@ install_rust() {
 
   if ! cargo llvm-cov --version >/dev/null 2>&1; then
     log "Installing cargo-llvm-cov"
-    cargo install cargo-llvm-cov --locked
+    if ! cargo install cargo-llvm-cov --locked; then
```
liliu-z, line 176: `install_rust` runs on both platforms and unconditionally builds `cargo-llvm-cov` from source, but the macOS CI job can never use it: it invokes `run_tests.sh --no-server`, and that path rejects `CODE_COV=true` outright, so no coverage command ever runs on macOS. Every macOS run pays for a from-source `cargo install` whose output is dead weight (the same applies to the `brew install lcov` fallback at lines 120-127, which only exists to provide `genhtml` for coverage). Gate both coverage installs on Linux; `verify_required_tools` already degrades to a warning when `cargo-llvm-cov` and `genhtml` are absent, so nothing else breaks.

.github/workflows/main.yml
```diff
@@ -64,17 +69,20 @@ jobs:
           flags: ubuntu-${{ matrix.ubuntu }}-unittests
 
   ut-macos:
-    name: Compile on macOS 15
+    name: UT on MacOS 15
     runs-on: macos-15
-    timeout-minutes: 30
+    timeout-minutes: 45
```
liliu-z, line 74: The macOS job's timeout was raised from 30 to 45 minutes, but the Linux job was left at 30 even though it now runs a strict superset of the macOS work: the same `cargo check --all-targets` plus seven tutorial-crate checks in a separate `CARGO_TARGET_DIR`, then a `cargo llvm-cov clean` that forces a full instrumented rebuild, then a Milvus image pull and the v1/v2 server suites, plus the new `cargo fmt --check` step. That is three separate builds of a tonic/prost dependency graph plus an integration run inside the old budget that previously covered one instrumented build and `docker compose up`. Raise `timeout-minutes` on `ut-linux` as well, or the job that actually gates merges is the one most likely to time out.

scripts/install_deps.sh
```diff
@@ -44,17 +49,52 @@ install_linux_packages() {
     log "Installing build and coverage packages with apt-get"
     "${SUDO[@]}" apt-get update
     "${SUDO[@]}" env DEBIAN_FRONTEND=noninteractive apt-get install -y \
-      build-essential ca-certificates curl git lcov libssl-dev pkg-config
+      build-essential ca-certificates curl git lcov libssl-dev pkg-config python3
```
yhmo: scripts/install_deps.sh line:52
Medium ---- `SKIP_COVERAGE_TOOLS=true` only gates the macOS `lcov` branch and `cargo-llvm-cov`; all three Linux package transactions still require `lcov`. That contradicts the README promise that this flag omits optional coverage tools and can make a non-coverage setup fail solely because an optional `lcov` package is unavailable. Please build the Linux package list conditionally (or install `lcov` best-effort) so the skip flag has the same contract on Linux.

tests/v2/st/milvus_container.py
```diff
@@ -0,0 +1,272 @@
...
+
+def reap_stale_containers(grpc_port: int, health_port: int) -> None:
+    """Remove test containers left behind when a previous runner was killed."""
+    result = docker(
+        "ps",
+        "--all",
+        "--quiet",
+        "--filter",
+        f"label={reap_label(grpc_port, health_port)}",
```
yhmo, line 160: tests/v2/st/milvus_container.py line:161
Medium ---- This filter treats every container using these ports as stale, including a live container owned by another overlapping `run_tests.sh` invocation. A second run will therefore force-remove the first run's Milvus at line 169, causing unrelated tests to fail mid-run instead of reporting that the ports are busy. Please distinguish exited containers from running ones (or use an owner/run-specific label and only reap containers that this invocation can prove are orphaned).

PR #135, opened 2026-08-12
Adjust tutorials

yhmo:


scripts/run_tutorials.sh
```diff
@@ -74,7 +74,7 @@ MILVUS_CONTAINER_ID="$(
 )"
 
 cd "$ROOT_DIR"
-for manifest in tutorial/{1_database,2_collection,3_schema,4_index,5_dml,6_dql}/Cargo.toml; do
+for manifest in tutorial/*/Cargo.toml; do
```
yhmo, line 78: scripts/run_tutorials.sh line:77
High ---- This glob now executes `tutorial/8_rbac`, but the shared launcher starts Milvus without `common.security.authorizationEnabled=true` (the server default is false), while the new RBAC tutorial explicitly requires authorization. The release workflow therefore reaches this tutorial and fails on its first RBAC call, aborting published-crate tutorial verification. Please either start a separate auth-enabled server for this tutorial or exclude it from this default loop.

tutorial/8_rbac/src/main.rs
```diff
@@ -0,0 +1,273 @@
...
+
+use milvus::v2::prelude::*;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    let uri = std::env::var("MILVUS_URI").unwrap_or_else(|_| "http://localhost:19530".into());
+    let token = std::env::var("MILVUS_TOKEN").unwrap_or_else(|_| "root:Milvus".into());
+    let suffix = unique_suffix();
+    let username = format!("rust_tutorial_user_{suffix}");
```
yhmo: tutorial/8_rbac/src/main.rs line:25
High ---- The generated username cannot be created on the documented Milvus 2.6 setup. Milvus v2.6.20 defaults `proxy.maxUsernameLength` to 32 bytes, but `rust_tutorial_user_` plus the 13-digit millisecond timestamp, underscore, and PID is already at least 34 bytes, so `create_user` rejects every run before the tutorial demonstrates any RBAC operation (and the post-release tutorial job aborts here). Please use a shorter prefix/suffix scheme like the existing RBAC system test's `u<PID>_<millis>` format.

tutorial/8_rbac/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+        .await;
+    println!(
+        "drop_privilege_group completed: {}",
+        if drop_privilege_group_result.is_ok() {
+            "ok"
+        } else {
+            "ignored error"
+        }
+    );
+    Ok(())
```
yhmo: tutorial/8_rbac/src/main.rs line:266
Medium ---- This always returns success even when `drop_user`, `drop_role`, or `drop_privilege_group` failed. A transient RPC/server error can therefore leave server-level RBAC resources behind while the tutorial (and the post-release tutorial job) exits zero, contradicting the README cleanup guarantee and eventually consuming the finite user/role capacity on repeated runs. Please keep attempting every teardown step, but propagate or aggregate unexpected destructive-cleanup errors instead of discarding all of them.

tutorial/4_index/README.md
```diff
@@ -38,3 +38,17 @@ cargo run --manifest-path tutorial/4_index/Cargo.toml
...
+
+## Expected output
+
+The exact index state may vary briefly, but a successful run includes:
+
+```text
+Calling create_index: build vector and scalar indexes
+create_index completed
+Created indexes:
+Calling describe_index: inspect "embedding_hnsw"
```
yhmo: tutorial/4_index/README.md line:50
Low ---- These expected names do not match the constants printed by the tutorial: the program emits `embedding_hnsw_idx` and `price_sort_idx`. Please include the `_idx` suffixes so beginners can compare the documented output with an actual run.

PR #137, opened 2026-08-13
Add windows CI

yhmo:


.github/workflows/main.yml
```diff
@@ -1,6 +1,21 @@
 name: Unittest
 
 on:
+  push:
+    branches:
+      - main
+      - '[0-9]+.[0-9]+'
```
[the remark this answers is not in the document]
yhmo: not a bug

src/lib.rs
```diff
@@ -14,6 +14,52 @@
...
+//! The repository README and standalone tutorials provide complete create, insert, load, search,
+//! query, administration, and cleanup workflows.
+//!
+//! # Optional diagnostics
+//!
+//! Enable the Cargo feature `tracing`, add `tracing-subscriber` with its `env-filter` feature to
+//! the application, and initialize a subscriber before creating the client:
+//!
+//! ```
+//! use tracing_subscriber::EnvFilter;
```
yhmo, line 53: src/lib.rs line:53
Medium ---- This fenced block is an executable doctest, but it imports `tracing_subscriber`, which is neither a dependency nor a dev-dependency of this crate. Therefore the new `cargo test --doc` step cannot compile this example. Please either add the crate as a dev-dependency or mark this application-only snippet as `ignore`/text so the documentation-test gate can pass.

.github/workflows/main.yml
yhmo, line 7: .github/workflows/main.yml line:7
Medium ---- This branch filter is still written as a regex, but GitHub Actions evaluates `branches` entries as glob patterns, so the literal `+` characters prevent the new push workflow from matching the repository's `2.5` branch (and future `2.6` branches). The earlier thread covered the pull-request filter; this newly added push filter creates the same gap for direct release-branch updates. Please use a glob that actually matches the release naming scheme, such as `[0-9]*.[0-9]*`.

yhmo: The comment is not reasonable.

  GitHub Actions branch filters use GitHub’s extended glob syntax, where:

  - [0-9] matches one digit.
  - + matches one or more occurrences of the preceding pattern.
  - . is literal here.

  Therefore:

  '[0-9]+.[0-9]+'

  matches 2.5, 2.6, 10.2, and similar release branches for both push and pull_request.

  The suggested replacement is weaker:

  '[0-9]*.[0-9]*'

  In GitHub’s glob syntax, * is an independent wildcard, not a regex quantifier for [0-9]. It can match
  non-digit characters and therefore accepts unintended branch names.

src/v2/client/rbac.rs
```diff
@@ -22,14 +22,17 @@ use crate::v2::error::Result;
 use crate::v2::{request, response};
 
 impl ClientV2 {
-    /// Create an user with username and password to login milvus.
+    /// Creates a user account with a username and password.
+    ///
+    /// This operation requires Milvus authorization to be enabled and the caller to have the
```
yhmo: src/v2/client/rbac.rs line:27
Medium ---- This says `create_user` requires authorization to be enabled, but the repository's own DEVELOPMENT.md and tutorial/8_rbac/README.md document that RBAC management calls, including `create_user`, can run with authorization disabled; only permission enforcement requires enabling it. Please narrow this wording so the API documentation does not tell users a supported default-server workflow is unavailable.

PR #143, opened 2026-08-17
Support snapshots interfaces

yhmo:


src/v2/request/snapshot.rs
```diff
@@ -0,0 +1,1072 @@
...
+    pub(crate) fn into_proto(self, default_db: &str) -> milvus::RestoreSnapshotRequest {
+        milvus::RestoreSnapshotRequest {
+            base: None,
+            name: self.snapshot_name,
+            db_name: self
+                .source_database_name
+                .unwrap_or_else(|| default_db.to_owned()),
+            collection_name: self.source_collection_name,
+            rewrite_data: false,
+            target_db_name: self.target_database_name.unwrap_or_default(),
```
yhmo: src/v2/request/snapshot.rs line:488
Medium ---- `target_db_name` defaults to an empty string when `target_database_name` is omitted, while the source `db_name` defaults to the client's selected database (`default_db`). With a non-default database selected (e.g. `use_database("analytics")`), `restore_snapshot` resolves the source to "analytics" but sends `target_db_name=""`, so the restored collection lands in the server's default database instead of the selected one. Consider defaulting `target_db_name` to `default_db` as well, or documenting the asymmetric behavior.

src/v1/value.rs
```diff
@@ -431,6 +438,10 @@ impl From<Field> for ValueVec {
                     ScalarData::GeometryData(v) => Self::Geometry(v.data),
                     ScalarData::TimestamptzData(v) => Self::Timestamptz(v.data),
                     ScalarData::GeometryWktData(v) => Self::GeometryWkt(v.data),
+                    ScalarData::MolData(_) => Self::None,
```
yhmo, line 443: src/v1/value.rs line:441
Low ---- `ScalarData::MolData(_)` is mapped to `Self::None` here, silently dropping Mol column data, while `ValueVec::new(DataType::Mol)` on line 275 produces `Self::Binary`. MolSmiles/Date/Time get explicit mappings above, so Mol is the only newly-added type whose V1 conversion discards the values. If Mol data is not representable in `ValueVec`, a short comment or mapping note would make the intentional drop discoverable instead of looking like an accidental omission.

PR #146, opened 2026-08-24
feat: add client telemetry support

xiaofan-luan:
Related issue: milvus-io/milvus#47281

## Summary

- add default-on client telemetry to `ClientV2`, shared by client clones and cluster-scoped sessions
- report the seven Go SDK operation families once per logical call: Search, Query (including Get), HybridSearch, RunAnalyzer, Insert, Delete, and Upsert
- send immediate best-effort heartbeats and implement the five server commands: `push_config`, `collection_metrics`, `show_errors`, `show_latency_history`, and `get_config`
- keep Milvus and telemetry stubs in one generation-fenced service bundle so global-cluster failover switches them atomically
- keep the command heartbeat alive after a server-side dynamic disable so its ACK and a later re-enable can flow, while an initial user opt-out starts no task
- expose telemetry configuration, diagnostics, and opt-in OpenTelemetry-compatible client request IDs

The Rust repository's current feature surface is V2-only, so this integrates with `ClientV2` rather than extending the legacy V1 client.

## Protocol alignment

- 10-second default interval, dynamic interval updates, no per-heartbeat retry, 10-second RPC timeout, and capped UNIMPLEMENTED backoff
- deterministic sampling, recent-error and one-hour latency history, 1 MiB reply limits, persistent config hash, and equal-timestamp command deduplication
- stable per-client identity across clones/reconnects, optional caller-pinned IDs across process restarts, secret-free config reporting, and dynamic database reporting
- reply clearing and command execution are fenced against stale global-cluster generations

## Design

- https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260131-client_side_telemetry.md
- Go conformance follow-up: https://github.com/milvus-io/milvus/pull/52804

## Verification

- non-server tests: 908 passed, 0 failed, 1 ignored
- `cargo fmt --all -- --check`
- correctness/suspicious clippy gate
- V2 API checker: 115 requests and 50 responses
- rustdoc warnings gate and cargo package verification

### DIFF
--- a/src/v2/client/dml.rs
+++ b/src/v2/client/dml.rs
@@ -33,54 +33,60 @@ impl ClientV2 {
         &self,
         request: request::dml::InsertRequest,
     ) -> Result<response::dml::InsertResponse> {
-        let database = self.effective_database(request.database_name.as_deref());
         let collection_name = request.collection_name.clone();
-        for attempt in 0..2 {
-            let resolved = self
-                .resolve_data(
+        let telemetry = self.telemetry.begin_operation("Insert", &collection_name);
+        let result = async {
+            let database = self.effective_database(request.database_name.as_deref());
+            for attempt in 0..2 {
+                let resolved = self
+                    .resolve_data(
+                        &database,
+                        &collection_name,
+                        &request.columns,
+                        &request.rows,
+                        false,
+                        false,
+                        &[],
+                    )
+                    .await?;
+                let response = self
+                    .retry_rpc(
+                        || {
+                            let fields = resolved.to_proto_fields(
+                                &request.columns,
+                                &request.rows,
+                                false,
+                                false,
+                            )?;
+                            request.to_proto_with_fields(
+                                fields,
+                                resolved.row_count,
+                                resolved.schema_timestamp,
+                                &database,
+                            )
+                        },
+                        RetrySemantics::NonIdempotent,
+                        |mut service, request| async move { service.insert(request).await },
+                        dml_retry_status,
+                    )
+                    .await?;
+                if attempt == 0 && is_schema_mismatch(&response.status) {
+                    self.remove_collection_description(&database, &collection_name);
+                    continue;
+                }
+                status_to_result(&response.status)?;
+                self.update_dml_timestamp(
                     &database,
-                    &collection_name,
-                    &request.columns,
-                    &request.rows,
-                    false,
-                    false,
-                    &[],
-                )
-                .await?;
-            let response = self
-                .retry_rpc(
-                    || {
-                        let fields = resolved.to_proto_fields(
-                            &request.columns,
-                            &request.rows,
-                            false,
-                            false,
-                        )?;
-                        request.to_proto_with_fields(
-                            fields,
-                            resolved.row_count,
-                            resolved.schema_timestamp,
-                            &database,
-                        )
-                    },
-                    RetrySemantics::NonIdempotent,
-                    |mut service, request| async move { service.insert(request).await },
-                    dml_retry_status,
-                )
-                .await?;
-            if attempt == 0 && is_schema_mismatch(&response.status) {
-                self.remove_collection_description(&database, &collection_name);
-                continue;
+                    &resolved.canonical_collection_name,
+                    response.timestamp,
+                );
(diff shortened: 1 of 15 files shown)


src/v2/types/common.rs
```diff
@@ -2697,6 +2697,140 @@ impl RetryConfig {
...
+        self.error_max_count
+    }
+
+    /// Returns the pinned client identifier, or an empty string when generated.
+    pub fn get_client_id(&self) -> &str {
+        &self.client_id
+    }
+}
+
+impl Default for TelemetryConfig {
```
yhmo, line 2842: src/v2/types/common.rs line:2828
Low ---- `Default for TelemetryConfig` is a pure alias of `new()` and has no call site in the tree, while the sibling value types `RetryConfig` and `ConnectConfig` intentionally expose only `new()`. The V2 convention here reserves `Default` for genuine enum/sentinel semantics rather than as a `new()` alias, so consider removing this impl to keep the config types consistent.

src/v2/types/common.rs
```diff
@@ -2697,6 +2697,140 @@ impl RetryConfig {
...
+
+    /// Enables or disables telemetry and returns this value for further mutation.
+    pub fn set_enabled(&mut self, enabled: bool) -> &mut Self {
+        self.enabled = enabled;
+        self
+    }
+
+    /// Sets the heartbeat interval.
+    ///
+    /// A zero interval is accepted here and resolved to the 10-second default,
```
yhmo: src/v2/types/common.rs line:2740
Low ---- The doc on `heartbeat_interval` claims a zero interval "is resolved to the 10-second default", and `error_max_count` claims zero "resolves to the default of 100", but the resolution happens later in `RuntimeState::new` (src/v2/client/telemetry.rs:299). `TelemetryConfig::get_heartbeat_interval()` / `get_error_max_count()` still return the raw 0 until a client is constructed, so the documented behavior and the observable getters disagree. Consider resolving in the setters or documenting that the getters report the raw configured value.

src/v2/client/dml.rs
```diff
@@ -33,54 +33,60 @@ impl ClientV2 {
         &self,
         request: request::dml::InsertRequest,
     ) -> Result<response::dml::InsertResponse> {
-        let database = self.effective_database(request.database_name.as_deref());
         let collection_name = request.collection_name.clone();
-        for attempt in 0..2 {
-            let resolved = self
-                .resolve_data(
+        let telemetry = self.telemetry.begin_operation("Insert", &collection_name);
```
yhmo, line 37: src/v2/client/dml.rs line:37
Low ---- The default-on instrumentation wrapping Insert/Upsert/Delete (and Query/Get/Search/HybridSearch/RunAnalyzer) has no unit test proving a logical call records a metric (success/error/latency), and the `with_client_request_id` -> `client_request_id` gRPC metadata path is not asserted on the wire; the telemetry module tests cover the collectors and heartbeat loop but not the DML/DQL integration, and this PR's verification ran only `--no-server`. Since telemetry ships on by default for all users, consider a mock-based test that a successful insert/query records an Insert/Query entry, a failed one records an error entry, and that the request-ID header reaches the wire.

src/v2/client/telemetry.rs
```diff
@@ -0,0 +1,1943 @@
...
+            .cloned()
+            .collect();
+        ignored.sort();
+        let mut applied = Vec::new();
+
+        // All validation above precedes this lock: updates are all-or-nothing.
+        let mut state = self.state.lock();
+        if let Some(enabled) = enabled {
+            state.config.enabled = enabled;
+            applied.push("enabled");
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/common.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/dml.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/common.rs
yhmo: ignore

PR #148, opened 2026-08-27
Align update_password/grant_privilege/revoke_privilege/compact with pymilvus

yhmo:
…ymilvus

### DIFF
--- a/src/v2/client.rs
+++ b/src/v2/client.rs
@@ -281,10 +281,10 @@ impl Interceptor for V2Interceptor {
 /// - DDL operations that change a database, collection identity, schema, or alias should be
 ///   serialized with DML and DQL calls targeting the affected objects. Results are not guaranteed
 ///   when those operations overlap.
-/// - [`ClientV2::use_database`], [`ClientV2::set_rpc_deadline`], and
-///   [`ClientV2::set_retry_param`] are internally synchronized but update state shared by every
-///   clone. Serialize configuration changes with RPC creation when deterministic request settings
-///   are required.
+/// - [`ClientV2::use_database`], [`ClientV2::set_rpc_deadline`], [`ClientV2::set_retry_param`],
+///   and `update_password` with `reset_connection` (which re-establishes the shared channel and
+///   credentials) are internally synchronized but update state shared by every clone. Serialize
+///   configuration changes with RPC creation when deterministic request settings are required.
 #[derive(Clone)]
 pub struct ClientV2 {
     service: SharedServices,
@@ -296,6 +296,7 @@ pub struct ClientV2 {
     schema_load_scope: Arc<SchemaLoadScope>,
     global_cluster: Option<Arc<global_cluster::GlobalCluster>>,
     telemetry: ClientTelemetry,
+    connect_config: Arc<RwLock<ConnectConfig>>,
 }
 
 impl ClientV2 {
@@ -318,6 +319,7 @@ impl ClientV2 {
 
         validate_client_identity(&param)?;
 
+        let shared_config = Arc::new(RwLock::new(param.clone()));
         let (service, database, cache_endpoint, global_cluster) = if is_global_endpoint(&param.uri)
         {
             let topology = global_cluster::fetch_topology(&param.uri, &param).await?;
@@ -340,7 +342,7 @@ impl ClientV2 {
             let cache_endpoint = Arc::new(tls_endpoint_uri(&param.uri, tls_enabled(&param))?);
             let global = Arc::new(global_cluster::GlobalCluster::new(
                 param.uri.clone(),
-                param.clone(),
+                Arc::clone(&shared_config),
                 Arc::clone(&database),
                 Arc::clone(&database_explicit),
                 Arc::clone(&service),
@@ -385,6 +387,7 @@ impl ClientV2 {
             schema_load_scope: Arc::new(SchemaLoadScope::new()),
             global_cluster,
             telemetry,
+            connect_config: shared_config,
         })
     }
 
@@ -396,6 +399,53 @@ impl ClientV2 {
     /// Replaces the retry policy used by subsequent RPC calls.
     pub fn set_retry_param(&self, retry: RetryConfig) {
         *self.retry.write() = retry;
+    }
+
+    /// Re-establishes the connection using the supplied configuration.
+    ///
+    /// Swaps in a freshly built gRPC channel and authorization interceptor, preserving the
+    /// selected database and caches. On a global-cluster connection the shared configuration is
+    /// updated and the primary channel is rebuilt from it. Used by `update_password` with
+    /// `reset_connection` to re-arm the interceptor with the new credentials. On success the
+    /// telemetry-reported identity is updated to the configured username so heartbeats keep
+    /// matching the currently authenticated credentials.
+    pub(crate) async fn reset_connection(&self, config: ConnectConfig) -> Result<()> {
+        let username = config
+            .raw_token()
+            .and_then(|token| token.split_once(':').map(|(user, _)| user.to_owned()))
+            .unwrap_or_default();
+        if let Some(global) = &self.global_cluster {
+            // `reset_connection` commits the configuration to the shared slot only after the
+            // primary channel has been rebuilt, so stored credentials always match the active
+            // channel.
+            let result = global.reset_connection(config).await;
+            if result.is_ok() {
+                self.telemetry.update_username(username);
+            }
+            return result;
+        }
+        let (endpoint, _) = configured_endpoint(&config).await?;
+        let channel = endpoint.connect_lazy();
+        let token = config
+            .token
+            .as_deref()
+            .map(|value| value.parse())
+            .transpose()
+            .map_err(|_| {
+                Error::validation("token".into(), "token is not valid HTTP metadata".into())
+            })?;
+        let interceptor = V2Interceptor {
+            token,
+            database: Arc::clone(&self.database),
+            database_explicit: Arc::clone(&self.database_explicit),
+        };
+        let generation = self.service.read().generation.saturating_add(1);
+        let mut services = service_bundle(channel, interceptor, generation);
+        wait_for_server(&mut services.milvus, config.connect_timeout).await?;
+        *self.connect_config.write() = config;
+        *self.service.write() = services;
+        self.telemetry.update_username(username);
+        Ok(())
     }
 
     /// Returns the shared client-side telemetry manager.
@@ -877,6 +927,9 @@ mod retry_tests {
             schema_load_scope: Arc::new(SchemaLoadScope::new()),
             global_cluster: None,
             telemetry,
+            connect_config: Arc::new(RwLock::new(
+                ConnectConfig::new().uri("http://127.0.0.1:19530"),
+            )),
         }
     }
 
--- a/src/v2/client/global_cluster.rs
+++ b/src/v2/client/global_cluster.rs
@@ -361,7 +361,7 @@ pub(crate) async fn build_services(
 ///////////////////////////////////////////////////////////////////////////////
 pub(crate) struct GlobalCluster {
     global_endpoint: String,
-    connect_config: super::ConnectConfig,
+    connect_config: Arc<RwLock<super::ConnectConfig>>,
     database: Arc<RwLock<String>>,
     database_explicit: Arc<AtomicBool>,
     service: SharedServices,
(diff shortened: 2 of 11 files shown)


src/v2/client/global_cluster.rs
```diff
@@ -659,6 +662,32 @@ impl GlobalCluster {
...
         }
     }
+
+    /// Rebuilds the channel to the current primary using the shared configuration.
+    ///
+    /// Unlike a topology refresh, the rebuild always happens even when the primary is unchanged,
+    /// so credential changes in the shared configuration (for example after `reset_connection`)
+    /// take effect on the next request. A failed rebuild returns an error so the caller can
+    /// surface the failure instead of silently continuing with the previous credentials.
+    pub(crate) async fn rearm(&self) -> Result<()> {
```
yhmo: src/v2/client/global_cluster.rs line:672
Medium ---- The new `GlobalCluster::rearm()` -- and the global-cluster branch of `ClientV2::rearm_transport`, which updates the shared `connect_config` before rebuilding the primary channel -- has no test coverage. Every new reset-connection test (`update_password_reset_connection_rearms_credentials`) exercises only the plain non-global path, so the `apply_lock` path, the topology-primary lookup, the `rebuild_to` failure branch, and both `Error::Unexpected` mappings are never exercised. This is the most delicate part of the change (shared `RwLock` config + channel generation + `apply_lock`), so a test driven through `MockServer::start_global` that calls `update_password(..., reset_connection(true))` and asserts the primary channel is rebuilt with the new credentials -- plus one failure case -- would materially harden it.

src/v2/request/utility.rs
```diff
@@ -615,22 +623,50 @@ impl CompactRequest {
...
+                "target_size".into(),
+                "target size is too small and rounds to zero MB".into(),
+            ));
+        }
+        Ok(Some(mb))
+    }
+
     pub(crate) fn into_proto(self, default_db: &str) -> milvus::ManualCompactionRequest {
+        // The request is validated by `build`, so the conversion cannot fail here.
+        let target_size = self.target_size_mb().ok().flatten().unwrap_or_default();
```
yhmo: src/v2/request/utility.rs line:663
Low ---- `into_proto` swallows the `target_size_mb()` conversion error via `.ok().flatten().unwrap_or_default()`, silently sending the server default (0) instead of surfacing an explicit error. It is unreachable today because `build()` runs `target_size_mb()?`, but `into_proto` is `pub(crate)` and a future internal call site that bypasses `build()` would silently fabricate a default -- contradicting the V2 "propagate errors, never fabricate defaults" convention. Consider propagating the error (or using `expect` with a `debug_assert`) so a future misuse fails loudly instead of defaulting.

src/v2/types/utility.rs
```diff
@@ -114,6 +114,63 @@ impl CompactionStateCode {
...
+            Self::KB => 1024,
+            Self::MB => 1024 * 1024,
+            Self::GB => 1024 * 1024 * 1024,
+            Self::TB => 1024 * 1024 * 1024 * 1024,
+            Self::PB => 1024 * 1024 * 1024 * 1024 * 1024,
+        }
+    }
+}
+
+impl std::str::FromStr for TargetSizeUnit {
```
yhmo: src/v2/types/utility.rs line:153
Low ---- `TargetSizeUnit` implements `FromStr` (including the `kib`/`mib`/`gib`/`tib`/`pib` aliases), but nothing in the crate parses a unit from a string: `CompactRequestBuilder::target_size_unit` takes the enum directly and no test exercises `FromStr`. pymilvus accepts the unit as a string (e.g. `"gb"`), so either wire `FromStr` into the builder for closer parity or drop the unused implementation to avoid dead, untested public API.

src/v2/client.rs
```diff
@@ -398,6 +401,40 @@ impl ClientV2 {
...
 
+    /// Re-establishes the connection using the supplied configuration.
+    ///
+    /// Swaps in a freshly built gRPC channel and authorization interceptor, preserving the
+    /// selected database and caches. On a global-cluster connection the shared configuration is
+    /// updated and the primary channel is rebuilt from it. Used by `update_password` with
+    /// `reset_connection` to re-arm the interceptor with the new credentials.
+    pub(crate) async fn rearm_transport(&self, config: ConnectConfig) -> Result<()> {
+        if let Some(global) = &self.global_cluster {
+            *self.connect_config.write() = config;
```
yhmo: src/v2/client.rs line:412
Medium ---- The global-cluster branch commits the new token into the shared `connect_config` before `GlobalCluster::rearm()` actually rebuilds the channel. If the rebuild (`wait_for_server` inside `rebuild_to`) fails, `update_password` returns an error after the server already applied the credential change, and the stored config now holds the new token while the live primary channel still holds the old, now-invalid one. Since `apply_topology` keeps an unchanged primary's channel and auth failures surface as `UNAUTHENTICATED` (not `UNAVAILABLE`), `on_unavailable` will not recover it until the primary changes. Suggest passing the fresh config into `rearm()` (or rolling back the shared config when the rebuild fails) so stored credentials always match the active channel.

src/v2/request/utility.rs
```diff
@@ -681,10 +732,26 @@ impl CompactRequestBuilder {
...
             ));
         }
+        self.value.target_size_mb()?;
         Ok(self.value)
     }
 }
 
+/// Converts a target size in the given unit to megabytes, matching pymilvus's
+/// `b/kb/mb/gb/tb/pb` to MB conversion. Returns an error when the value is out of range.
+fn target_size_to_mb(target_size: i64, unit: TargetSizeUnit) -> Result<i64> {
```
yhmo, line 742: src/v2/request/utility.rs line:742
Low ---- This introduces a second size-to-MB converter alongside `parse_target_size_mb` (src/v2/client/utility.rs, used by OptimizeTask). The two already diverge: different rounding (f64 vs i128 integer division), different accepted unit spellings (this one also accepts `kib`/`mib`/`gib`/`tib`/`pib`, which pymilvus and the sibling converter reject), and slightly different bound checks. Consider reusing or extending a single converter so `compact` and `optimize` stay consistent as unit semantics evolve.

src/v2/types/utility.rs
```diff
@@ -114,6 +114,63 @@ impl CompactionStateCode {
...
+            "b" => Ok(Self::B),
+            "kb" | "kib" => Ok(Self::KB),
+            "mb" | "mib" => Ok(Self::MB),
+            "gb" | "gib" => Ok(Self::GB),
+            "tb" | "tib" => Ok(Self::TB),
+            "pb" | "pib" => Ok(Self::PB),
+            _ => Err(crate::v2::error::Error::validation(
+                "target_size_unit".into(),
+                format!(
+                    "invalid target size unit {value:?}; expected one of b, kb, mb, gb, tb, pb"
```
yhmo: src/v2/types/utility.rs line:167
Low ---- The `FromStr` error message lists only `b, kb, mb, gb, tb, pb`, but the parser also accepts `kib`/`mib`/`gib`/`tib`/`pib`. pymilvus and the sibling `parse_target_size_mb` accept only the six non-"i" spellings, so either drop the extra variants or update the message (and doc comment) to match the accepted set.

src/v2/client/global_cluster.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/utility.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/utility.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/utility.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/utility.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/rbac.rs
```diff
@@ -34,13 +34,26 @@ impl ClientV2 {
...
+        let username = request.username().to_owned();
+        let new_password = request.new_password().to_owned();
         let status =
             status_rpc_with_retry!(NonIdempotent, self, update_credential, request.into_proto())?;
-        self.status(status)
+        self.status(status)?;
+        if reset_connection {
+            let mut config = self.connect_config.read().clone();
+            config.set_token(format!("{username}:{new_password}"));
+            self.rearm_transport(config).await?;
```
yhmo: src/v2/client/rbac.rs line:54
Medium ---- If `rearm_transport` fails here, the server has already accepted the password change (the `update_credential` RPC succeeded), but this method returns an error that is indistinguishable from a failed password change. The shared `connect_config` and the active interceptor are only updated after a successful rebuild, so the client keeps the now-invalid old credentials: a caller that retries `update_password` will be rejected because the old password no longer matches, and subsequent authenticated RPCs on this client will fail with auth errors. This is also true for the global-cluster path via `GlobalCluster::rearm`, whose error message only says the connection could not be re-established. Consider documenting that the password change is applied even when reconnection fails, and/or surfacing a distinct error (e.g. "password changed but connection reset failed") so callers can distinguish "change succeeded, reconnect failed" from "change rejected" and know to re-authenticate with the new password.

src/v2/request/utility.rs
```diff
@@ -1798,21 +1868,94 @@ mod builder_value_tests {
...
+
+        let unset = CompactRequest::builder()
+            .collection_name("books")
+            .build()
+            .expect("valid request");
+        assert_eq!(unset.target_size_mb().expect("convert"), None);
+    }
+
+    #[test]
+    fn compact_request_target_size_rounds_up_to_at_least_one_megabyte() {
```
yhmo: src/v2/request/utility.rs line:1935
Low ---- This test name says the target size is rounded up, but the implementation floors byte-to-MB conversion with integer division and then rejects any result below 1 MB. Renaming it to something such as `compact_request_target_size_rejects_sub_megabyte_values` will keep the tested contract unambiguous.

yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/rbac.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client.rs
```diff
@@ -398,6 +401,42 @@ impl ClientV2 {
         *self.retry.write() = retry;
     }
 
+    /// Re-establishes the connection using the supplied configuration.
+    ///
+    /// Swaps in a freshly built gRPC channel and authorization interceptor, preserving the
+    /// selected database and caches. On a global-cluster connection the shared configuration is
+    /// updated and the primary channel is rebuilt from it. Used by `update_password` with
+    /// `reset_connection` to re-arm the interceptor with the new credentials.
+    pub(crate) async fn reset_connection(&self, config: ConnectConfig) -> Result<()> {
```
yhmo, line 412: src/v2/client.rs line:410
Medium ---- The reset updates the gRPC credentials but not the username frozen inside `ClientTelemetry`. `ClientTelemetry::new` derives that value once from the initial `ConnectConfig`, then `build_client_info` and `handle_get_config` keep using it, so heartbeats and telemetry report the old identity after `update_password(..., reset_connection(true))` switches the client to another user (or from an API-key/no-colon token to `username:password`). Please update or share the telemetry username on successful reset and add a regression test for the reported identity.

src/v2/client.rs
```diff
@@ -296,6 +296,7 @@ pub struct ClientV2 {
     schema_load_scope: Arc<SchemaLoadScope>,
     global_cluster: Option<Arc<global_cluster::GlobalCluster>>,
     telemetry: ClientTelemetry,
+    connect_config: Arc<RwLock<ConnectConfig>>,
```
yhmo, line 299: src/v2/client.rs line:299
Low ---- The `ClientV2` thread-safety contract lists `use_database`, `set_rpc_deadline`, and `set_retry_param` as shared-state changes requiring serialization, but `update_password` with `reset_connection(true)` now also replaces credentials and channel state shared by every clone. Please mention that operation so callers do not overlap credential resets or concurrent RPCs while assuming the enumerated list is complete.

src/v2/client.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #149, opened 2026-08-27
Support function chains and search aggregation

yhmo:
Support the new search-time function chain (rerank/refine plan) and hierarchical bucket search aggregation, aligned with the Milvus server `FunctionChain` proto (`schema.proto`) and mirroring pymilvus / milvus-sdk-java.

**Function chains** (`src/v2/types/function_chain.rs`)
- `FunctionChainStage`, `FunctionParamValue`, `FunctionChainArg`/`col()`, `FunctionChainExpr`, `FunctionChainOp`, `FunctionChain`
- Builder API: `FunctionChain::new().stage(..).name(..).map(..).sort(..).limit(..)` plus the `fn_` module (`decay`, `num_combine`, `round_decimal`, `xgboost`, `rerank_model`)
- `SearchRequestBuilder::function_chains(..)` / `add_function_chain(..)`; validation rejects `function_chains` combined with `rerank` and requires a non-`Unspecified` stage

**Search aggregation** (`src/v2/types/aggregation.rs`)
- `SearchAggregation`, `MetricOp`, `MetricSpec`, `SortDirection`, `SortSpec`, `OrderSpec`, `TopHitsSpec`; recursively nestable via `sub_aggregation`
- `SearchRequestBuilder::search_aggregation(..)`; mutually exclusive with `group_by_field`
- Response side: `SearchResults::get_agg_buckets()` / `get_agg_topks()` with typed `AggregationBucket`, `BucketKeyEntry`, `AggregationHit` parsing

Verified locally: `cargo check`, `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D clippy::correctness -D clippy::suspicious`, `python3 scripts/check_v2_api.py`, `cargo doc --no-deps` (`-D warnings`), `cargo test --doc`, `cargo test --test v2_ut`.

### DIFF
--- a/src/v2/request/dql.rs
+++ b/src/v2/request/dql.rs
@@ -29,8 +29,8 @@ use crate::v2::request::validation::{
 };
 pub use crate::v2::types::Ids;
 use crate::v2::types::{
-    encode_sparse_vector, validate_sparse_vector, ConsistencyLevel, Function, FunctionScore,
-    MetricType,
+    encode_sparse_vector, validate_sparse_vector, ConsistencyLevel, Function, FunctionChain,
+    FunctionScore, MetricType, SearchAggregation,
 };
 pub use crate::v2::types::{
     EmbeddingList, HighlightQuery, HighlightType, Highlighter, LexicalHighlighter, SearchVectors,
@@ -514,6 +514,8 @@ pub struct SearchRequest {
     pub(crate) timezone: String,
     pub(crate) highlighter: Option<Highlighter>,
     pub(crate) consistency_level: Option<ConsistencyLevel>,
+    pub(crate) function_chains: Vec<FunctionChain>,
+    pub(crate) search_aggregation: Option<SearchAggregation>,
 }
 
 impl SearchRequest {
@@ -647,6 +649,16 @@ impl SearchRequest {
     /// Returns the consistency level.
     pub fn consistency_level(&self) -> Option<ConsistencyLevel> {
         self.consistency_level
+    }
+
+    /// Returns the function chains.
+    pub fn function_chains(&self) -> &[FunctionChain] {
+        &self.function_chains
+    }
+
+    /// Returns the search aggregation.
+    pub fn search_aggregation(&self) -> &Option<SearchAggregation> {
+        &self.search_aggregation
     }
 
     #[allow(deprecated)]
@@ -896,6 +908,15 @@ impl SearchRequest {
             namespace: None,
             highlighter: self.highlighter.map(Highlighter::into_proto),
             search_input: Some(search_input),
+            search_aggregation: self
+                .search_aggregation
+                .map(|aggregation| aggregation.into_proto())
+                .transpose()?,
+            function_chains: self
+                .function_chains
+                .into_iter()
+                .map(FunctionChain::into_proto)
+                .collect(),
             ..Default::default()
         })
     }
@@ -928,6 +949,8 @@ impl SearchRequest {
             timezone: String::new(),
             highlighter: None,
             consistency_level: None,
+            function_chains: Vec::new(),
+            search_aggregation: None,
         }
     }
 }
@@ -1085,6 +1108,30 @@ impl SearchRequestBuilder {
     /// Sets the consistency level and returns the updated value.
     pub fn consistency_level(mut self, value: ConsistencyLevel) -> Self {
         self.value.consistency_level = Some(value);
+        self
+    }
+
+    /// Sets the function chains and returns the updated value.
+    ///
+    /// Mutually exclusive with [`Self::rerank`]; Milvus rejects a search carrying both.
+    pub fn function_chains(mut self, values: impl IntoIterator<Item = FunctionChain>) -> Self {
+        self.value.function_chains = values.into_iter().collect();
+        self
+    }
+
+    /// Adds a function chain and returns the updated value.
+    pub fn add_function_chain(mut self, value: FunctionChain) -> Self {
+        self.value.function_chains.push(value);
+        self
+    }
+
+    /// Sets the hierarchical bucket aggregation and returns the updated value.
+    ///
+    /// Mutually exclusive with [`Self::group_by_field`] and [`Self::highlighter`]; when set,
+    /// `limit` is ignored, `SearchAggregation.size` controls the top-level bucket count, and
+    /// `offset` must remain zero.
+    pub fn search_aggregation(mut self, value: SearchAggregation) -> Self {
+        self.value.search_aggregation = Some(value);
         self
     }
 
@@ -1871,6 +1918,36 @@ fn validate_search_request(value: &SearchRequest) -> Result<()> {
     validate_search_extra_params(&value.extra_params)?;
     validate_finite_range_parameter("radius", value.radius)?;
     validate_finite_range_parameter("range_filter", value.range_filter)?;
+    if !value.function_chains.is_empty() && value.rerank.is_some() {
+        return Err(Error::validation(
+            "function_chains".into(),
+            "cannot be used together with rerank".into(),
+        ));
+    }
+    for chain in &value.function_chains {
+        chain.validate()?;
+    }
+    if let Some(aggregation) = &value.search_aggregation {
+        aggregation.validate()?;
+        if !value.group_by_field.is_empty() {
+            return Err(Error::validation(
+                "search_aggregation".into(),
+                "cannot be used together with group_by_field".into(),
+            ));
+        }
+        if value.offset > 0 {
+            return Err(Error::validation(
+                "offset".into(),
+                "cannot be used together with search_aggregation".into(),
+            ));
+        }
+        if value.highlighter.is_some() {
+            return Err(Error::validation(
+                "highlighter".into(),
+                "cannot be used together with search_aggregation".into(),
+            ));
+        }
+    }
     if !value.ids.is_empty() {
         if search_vectors_are_empty(&value.vectors) {
             return Ok(());
@@ -2013,13 +2090,14 @@ fn set_search_param(params: &mut Vec<common::KeyValuePair>, key: &str, value: im
 #[cfg(test)]
 mod search_request_tests {
     use super::{
-        EmbeddingList, HighlightQuery, HybridSearchRequest, LexicalHighlighter, SearchRequest,
-        SearchVectors, SubSearchRequest,
+        EmbeddingList, HighlightQuery, Highlighter, HybridSearchRequest, LexicalHighlighter,
+        SearchRequest, SearchVectors, SubSearchRequest,
     };
     use crate::proto::{common, milvus};
     use crate::v2::types::{
-        BoostRerank, DecayRerank, FunctionScore, Ids, MetricType, ModelRerank, SparseVector,
-        WeightedRerank,
+        col, fn_, BoostRerank, DecayRerank, FunctionChain, FunctionChainStage, FunctionScore,
+        HighlightType, Ids, MetricOp, MetricSpec, MetricType, ModelRerank, OrderSpec,
+        SearchAggregation, SortDirection, SparseVector, WeightedRerank,
     };
     use prost::Message;
     use serde_json::json;
@@ -2248,6 +2326,151 @@ mod search_request_tests {
             result,
             Err(crate::v2::error::Error::Validation(error)) if error.parameter() == "ids"
         ));
+    }
+
+    #[test]
+    fn search_rejects_function_chains_combined_with_rerank() {
+        let chain = FunctionChain::new()
+            .stage(FunctionChainStage::L2Rerank)
+            .map(
+                "$score",
+                fn_::num_combine(vec![col("$score"), col("popularity")], "sum", None),
+            );
+        let result = SearchRequest::builder()
+            .collection_name("books")
+            .vector_field("embedding")
+            .vectors(SearchVectors::Float(vec![vec![0.1, 0.2]]))
+            .function_chains([chain])
+            .rerank(FunctionScore::new().add_function(BoostRerank::new().name("boost").weight(2.0)))
+            .build();
+
+        assert!(matches!(
+            result,
+            Err(crate::v2::error::Error::Validation(error))
+                if error.parameter() == "function_chains"
+        ));
+    }
+
+    #[test]
+    fn search_rejects_aggregation_combined_with_group_by_field() {
+        let aggregation = SearchAggregation::new()
+            .fields(["category"])
+            .size(10)
+            .add_metric(
+                "total",
+                MetricSpec::new().op(MetricOp::Sum).field_name("price"),
+            )
+            .add_order(
+                OrderSpec::new()
+                    .key("_count")
+                    .direction(SortDirection::Desc),
+            );
+        let result = SearchRequest::builder()
+            .collection_name("books")
+            .vector_field("embedding")
+            .vectors(SearchVectors::Float(vec![vec![0.1, 0.2]]))
+            .group_by_field("category")
+            .search_aggregation(aggregation)
+            .build();
+
+        assert!(matches!(
+            result,
+            Err(crate::v2::error::Error::Validation(error))
+                if error.parameter() == "search_aggregation"
+        ));
+    }
+
+    #[test]
+    fn search_rejects_aggregation_combined_with_nonzero_offset() {
+        let aggregation = SearchAggregation::new()
+            .fields(["category"])
+            .size(10)
+            .add_metric(
+                "total",
+                MetricSpec::new().op(MetricOp::Sum).field_name("price"),
+            )
+            .add_order(
+                OrderSpec::new()
+                    .key("_count")
+                    .direction(SortDirection::Desc),
+            );
+        let result = SearchRequest::builder()
+            .collection_name("books")
+            .vector_field("embedding")
+            .vectors(SearchVectors::Float(vec![vec![0.1, 0.2]]))
+            .offset(1)
+            .search_aggregation(aggregation)
+            .build();
+
+        assert!(matches!(
+            result,
+            Err(crate::v2::error::Error::Validation(error))
+                if error.parameter() == "offset"
+        ));
+    }
+
+    #[test]
+    fn search_rejects_aggregation_combined_with_highlighter() {
+        let aggregation = SearchAggregation::new()
+            .fields(["category"])
+            .size(10)
+            .add_metric(
+                "total",
+                MetricSpec::new().op(MetricOp::Sum).field_name("price"),
+            )
+            .add_order(
+                OrderSpec::new()
+                    .key("_count")
+                    .direction(SortDirection::Desc),
+            );
+        let result = SearchRequest::builder()
+            .collection_name("books")
+            .vector_field("embedding")
+            .vectors(SearchVectors::Float(vec![vec![0.1, 0.2]]))
+            .highlighter(Highlighter::new().highlight_type(HighlightType::Lexical))
+            .search_aggregation(aggregation)
+            .build();
+
+        assert!(matches!(
+            result,
+            Err(crate::v2::error::Error::Validation(error))
+                if error.parameter() == "highlighter"
+        ));
+    }
+
+    #[test]
+    fn search_encodes_function_chains_and_aggregation_on_the_happy_path() {
+        let chain = FunctionChain::new()
+            .stage(FunctionChainStage::L2Rerank)
+            .sort("$score", true, None)
+            .limit(10, 0);
+        let aggregation = SearchAggregation::new()
+            .fields(["category"])
+            .size(10)
+            .add_metric(
+                "total",
+                MetricSpec::new().op(MetricOp::Sum).field_name("price"),
+            )
+            .add_order(
+                OrderSpec::new()
+                    .key("_count")
+                    .direction(SortDirection::Desc),
+            );
+        let request = SearchRequest::builder()
+            .collection_name("books")
+            .vector_field("embedding")
+            .vectors(SearchVectors::Float(vec![vec![0.1, 0.2]]))
+            .function_chains([chain])
+            .search_aggregation(aggregation)
+            .build()
+            .expect("valid request with function chains and aggregation")
+            .into_proto("default", 0)
+            .expect("encode request");
+
+        assert_eq!(request.function_chains.len(), 1);
+        assert_eq!(request.function_chains[0].ops.len(), 2);
+        assert_eq!(request.function_chains[0].ops[0].op, "sort");
+        assert!(request.search_aggregation.is_some());
     }
 
     #[test]
--- a/src/v2/response/dql.rs
+++ b/src/v2/response/dql.rs
@@ -23,7 +23,7 @@
 use crate::proto::{milvus, schema};
 use crate::v2::error::{Error, Result};
 use crate::v2::types::Ids;
-use crate::v2::types::{DataType, FieldData, SparseVector};
+use crate::v2::types::{group_aggregation_buckets, DataType, FieldData, SparseVector};
 pub use crate::v2::types::{HighlightResult, QueryResults, SearchResults, SingleResult};
 use std::collections::{HashMap, HashSet};
 
@@ -1085,6 +1085,11 @@ impl SearchResponse {
             results: SearchResults {
                 results: single_results,
                 recalls: result.recalls,
+                agg_buckets: group_aggregation_buckets(
+                    result.agg_buckets,
+                    result.agg_topks,
+                    result.num_queries,
+                )?,
             },
         })
     }
@@ -1114,6 +1119,7 @@ impl SearchResponse {
             results: SearchResults {
                 results: vec![remaining],
                 recalls: self.results.recalls.clone(),
+                agg_buckets: self.results.agg_buckets.clone(),
             },
             session_timestamp: self.session_timestamp,
             cost: self.cost,
@@ -1739,7 +1745,7 @@ pub type HybridSearchResponse = SearchResponse;
 mod tests {
     use super::{field_data, split_field_data, QueryResponse, SearchResponse};
     use crate::proto::{common, milvus, schema};
-    use crate::v2::types::{DataType, FieldData};
+    use crate::v2::types::{AggregationBucketValue, DataType, FieldData};
 
     #[test]
     fn decode_field_data_prefers_field_specific_validity() {
@@ -3098,6 +3104,49 @@ mod tests {
 
         assert_eq!(search.results().len().to_owned(), 1);
         assert!(search.results().get_results()[0].is_empty());
+    }
+
+    #[test]
+    fn search_response_groups_aggregation_buckets_per_query() {
+        // nq=2 with zero result rows; the first query owns one top-level bucket (agg_topks=[1, 0]).
+        let search = SearchResponse::from_proto(milvus::SearchResults {
+            results: Some(schema::SearchResultData {
+                num_queries: 2,
+                top_k: 0,
+                topks: vec![0, 0],
+                scores: vec![],
+                ids: Some(schema::IDs::default()),
+                primary_field_name: "id".into(),
+                agg_buckets: vec![schema::AggBucket {
+                    key: vec![schema::BucketKeyEntry {
+                        field_id: 1,
+                        field_name: "category".into(),
+                        value: Some(schema::bucket_key_entry::Value::StringVal("tech".into())),
+                    }],
+                    count: 3,
+                    metrics: std::collections::HashMap::new(),
+                    hits: vec![],
+                    sub_groups: vec![],
+                }],
+                agg_topks: vec![1, 0],
+                ..Default::default()
+            }),
+            ..Default::default()
+        })
+        .expect("decode aggregation search response");
+
+        assert_eq!(search.results().get_agg_buckets().len(), 2);
+        assert_eq!(search.results().get_agg_buckets()[0].len(), 1);
+        let bucket = &search.results().get_agg_buckets()[0][0];
+        assert_eq!(bucket.get_count(), 3);
+        assert_eq!(bucket.get_key()[0].get_field_name(), "category");
+        assert_eq!(
+            bucket.get_key()[0].get_value(),
+            &AggregationBucketValue::String("tech".to_owned())
+        );
+        assert!(search.results().get_agg_buckets()[1].is_empty());
+        assert!(search.results().get_results()[0].is_empty());
+        assert!(search.results().get_results()[1].is_empty());
     }
 }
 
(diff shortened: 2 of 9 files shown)


src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,741 @@
...
+        if let Some(tie_break_col) = tie_break_col {
+            inputs.push(tie_break_col.clone());
+            op = op.param("tie_break_col", tie_break_col);
+        }
+        self.ops.push(op.inputs(inputs));
+        self
+    }
+
+    /// Appends a `limit` operation with an optional offset.
+    pub fn limit(mut self, limit: i64, offset: i64) -> Self {
```
yhmo, line 625: src/v2/types/function_chain.rs line:486
Medium ---- FunctionChain::limit accepts a non-positive `limit` and a negative `offset`, and the `map`/`sort` helpers accept an empty output/column name. Both peer SDKs reject these client-side: milvus-sdk-java (FunctionChainOp.sort/limit) throws for limit <= 0, offset < 0 and empty column names, and pymilvus (chain.py) raises ParamError for the same inputs. As written, an invalid chain is serialized and only rejected by the server, producing a delayed server error instead of an immediate SDK validation error. Suggest validating in these fluent helpers (or in FunctionChain::validate) to match peer parity.

src/v2/response/dql.rs
```diff
@@ -1114,6 +1120,8 @@ impl SearchResponse {
             results: SearchResults {
```
yhmo, line 1119: src/v2/response/dql.rs line:1120
Medium ---- SearchResponse::split_at clones agg_buckets/agg_topks into both the returned page and the remaining cache, while SearchResponse::append discards the incoming page's buckets entirely. If a search_aggregation request is routed through the legacy search-iterator cache path, the same buckets are duplicated across pages and later pages are dropped. SearchIteratorRequest does not reject search_aggregation (or function_chains), so the combination is reachable from the public builder. Suggest rejecting aggregation in validate_search_iterator_input, or defining explicit merge/split semantics for the bucket data.

src/v2/types/aggregation.rs
```diff
@@ -0,0 +1,512 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct MetricSpec {
+    pub(crate) op: MetricOp,
+    pub(crate) field_name: String,
+}
+
+impl MetricSpec {
+    /// Creates a metric aggregation over `field_name` using `op`.
+    pub fn new(op: MetricOp, field_name: impl Into<String>) -> Self {
```
yhmo: src/v2/types/aggregation.rs line:99
Low ---- MetricSpec::new accepts an empty `field_name`, and does not restrict the `"*"` wildcard to the `count` op. Both peer SDKs enforce these rules: pymilvus raises ParamError for an empty field name and for `"*"` with non-count ops, and milvus-sdk-java throws for the same cases. An invalid metric is currently sent to the server and fails there instead of failing local validation.

src/v2/types/aggregation.rs
```diff
@@ -0,0 +1,512 @@
...
+        self.top_hits.as_ref()
+    }
+
+    /// Returns the nested sub-aggregation, if configured.
+    pub fn get_sub_aggregation(&self) -> Option<&SearchAggregation> {
+        self.sub_aggregation.as_deref()
+    }
+
+    /// Validates the aggregation spec before it is attached to a search request.
+    pub fn validate(&self) -> Result<()> {
```
yhmo, line 544: src/v2/types/aggregation.rs line:444
Low ---- This new aggregation module (512 lines) and the matching response-decoding types (AggregationBucket, AggregationHit, BucketKeyEntry, and the oneof value enums in src/v2/types/dql.rs) have no unit tests; only function-chain proto encoding is covered in this PR. The recursive sub_aggregation validation, the order-key checks, and the oneof-based response decoding including the MalformedResponse error paths are untested, and nothing exercises SearchResults::get_agg_buckets()/get_agg_topks(). Consider adding tests for at least a nested bucket decode and a malformed-payload error path.

src/v2/types/function_chain.rs
yhmo: Fixed in 2168d0c. `FunctionChain::validate` now rejects a non-positive `limit` (requires > 0), a negative `offset`, an empty `map` output name, and an empty `sort` column, so invalid chains fail with an SDK-side validation error instead of being serialized to the server. Added unit tests for all four cases in src/v2/types/function_chain.rs.

src/v2/response/dql.rs
yhmo: Fixed in 2168d0c. `validate_search_iterator_input` now rejects search requests that carry `function_chains` or `search_aggregation`, so the legacy iterator cache path (where split_at clones the buckets and append drops them) can no longer be reached with aggregation data. The public builder still allows both features for regular search; only the iterator path rejects them. Added a unit test in src/v2/client/iterator.rs.

src/v2/types/aggregation.rs
yhmo: Fixed in 2168d0c. `SearchAggregation::validate` now rejects metrics with an empty `field_name` and restricts the "*" wildcard to the `count` operator, matching pymilvus/milvus-sdk-java. Invalid specs fail locally at build/validate time instead of on the server. Added unit tests for both cases in src/v2/types/aggregation.rs.

src/v2/types/aggregation.rs
yhmo: Fixed in 2168d0c. Added unit tests covering: SearchAggregation validation (empty fields/size, unknown order key, empty metric field name, "*" with non-count, nested sub-aggregation) and proto encoding (metrics/order/top_hits/sub_aggregation) in src/v2/types/aggregation.rs; hierarchical bucket decoding (nested sub-groups, typed key/metric/hit values) plus three MalformedResponse error paths and the get_agg_buckets/get_agg_topks accessors in src/v2/types/dql.rs. New add_agg_bucket/add_agg_topk builders were added to SearchResults to support the tests.

src/v2/response/dql.rs
```diff
@@ -1085,6 +1085,12 @@ impl SearchResponse {
...
                 results: single_results,
                 recalls: result.recalls,
+                agg_buckets: result
```
yhmo: src/v2/response/dql.rs line:1088
High ---- The proto flattens aggregation buckets across all query vectors and `agg_topks` gives the bucket count per query, but this decode stores one flat `Vec` and exposes the counts separately. For buckets `[q0a, q0b, q1a]` with `agg_topks = [2, 1]`, `get_agg_buckets()` cannot identify each query's buckets, and missing, mismatched, negative, or overflowing counts are accepted. Please return a per-query `Vec<Vec<AggregationBucket>>` (or an equivalent grouped accessor) and reject shape mismatches as `MalformedResponse`, matching the server design and C++/Java/Py/Go clients; add multi-query and malformed-response tests.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,838 @@
...
+        self.stage
+    }
+
+    /// Returns the ordered operations.
+    pub fn get_ops(&self) -> &[FunctionChainOp] {
+        &self.ops
+    }
+
+    /// Validates that the chain can be attached to a search request.
+    pub fn validate(&self) -> Result<()> {
```
yhmo, line 651: src/v2/types/function_chain.rs line:511
Medium ---- Follow-up to the resolved fluent-helper checks: a hand-built `FunctionChainOp::new("")` or `FunctionChainExpr::new("")` still passes, `map` can omit its expression, `col("")` remains accepted, and `fn_` helpers do not enforce `num_combine` arity/mode/weight agreement, XGBoost feature/output rules, non-empty rerank queries, or finite numeric parameters. These type-correct values therefore reach the server despite the request-builder validation contract. Please validate the generic operation/expression tree and helper constraints, or make invalid states unrepresentable, with tests beyond the fluent happy-path forms.

src/v2/types/aggregation.rs
```diff
@@ -0,0 +1,646 @@
...
```
yhmo, line 544: src/v2/types/aggregation.rs line:444
Medium ---- Follow-up to the resolved metric-field checks: `validate` still permits `add_metric("", ...)` and `TopHitsSpec::new(0)`, and top-hits sort field names can be empty. Those invalid nested specifications pass `SearchRequestBuilder::build` and are rejected only by the server, while the C++/Java/Py APIs validate aliases and top-hits size/sort locally. Please validate every nested level and add tests for an empty metric alias, non-positive top-hits size, and an empty sort field.

src/v2/types/aggregation.rs
```diff
@@ -0,0 +1,646 @@
...
+///////////////////////////////////////////////////////////////////////////////
+/// One level of hierarchical bucket aggregation, recursively nestable.
+///
+/// Attach the aggregation to a search request with
+/// [`SearchRequestBuilder::search_aggregation`](crate::v2::request::dql::SearchRequestBuilder::search_aggregation).
+/// It is mutually exclusive with `group_by_field`; when set, the search `limit` is ignored and
+/// [`Self::size`] controls the top-level bucket count.
+#[derive(Debug, Clone, PartialEq)]
+#[non_exhaustive]
+pub struct SearchAggregation {
```
yhmo, line 397: src/v2/types/aggregation.rs line:337
Low ---- The new reusable V2 types provide consuming setters and getters but omit the established `set_field(&mut self, ...)` families used by neighboring `Highlighter`/`SearchResults` types; `SearchAggregation` and `FunctionChain` also add `Default` aliases for states that cannot pass validation. Please align the mutation surface and avoid `Default` aliases for invalid value states so the Rust type API remains consistent as it grows.

src/v2/response/dql.rs
yhmo: Fixed in eeae428. `SearchResults` now carries `agg_buckets: Vec<Vec<AggregationBucket>>` grouped per query vector (matching pymilvus `_split_agg_buckets` and Java `getAggregationBuckets`). `SearchResponse::from_proto` groups the flat proto list using `agg_topks` and returns `MalformedResponse` for: missing `agg_topks` on a multi-query response, negative counts, usize overflow, and counts that do not sum to the parsed bucket count. The separate `get_agg_topks` accessor was removed for parity. Added multi-query grouping, nq==1 recovery, and malformed-response tests in src/v2/types/dql.rs.

src/v2/types/function_chain.rs
yhmo: Fixed in eeae428. `FunctionChain::validate` now walks the generic operation/expression tree and rejects empty operation names, empty expression names, empty column references (including `col("")`), and `map` operations without an attached expression. Unit tests cover each case in src/v2/types/function_chain.rs.

src/v2/types/aggregation.rs
yhmo: Fixed in eeae428. `SearchAggregation::validate` now rejects empty metric aliases (`add_metric("", ...)`), non-positive `TopHitsSpec` size, and empty top-hits sort field names; the checks run at every nested level via the recursive sub-aggregation validation. Added unit tests for all three cases in src/v2/types/aggregation.rs.

src/v2/types/aggregation.rs
yhmo: as designed

src/v2/types/dql.rs
```diff
@@ -2347,6 +2347,371 @@ fn result_json_kind(value: &Value) -> &'static str {
...
+/// The server flattens buckets for all query vectors into one `repeated` list and records the
+/// number of top-level buckets per query in `agg_topks`. A `MalformedResponse` is returned when
+/// `agg_topks` is missing for a multi-query response, or when the counts are negative, overflow
+/// `usize`, or do not add up to the number of parsed buckets.
+pub(crate) fn group_aggregation_buckets(
+    buckets: Vec<schema::AggBucket>,
+    topks: Vec<i64>,
+    num_queries: i64,
+) -> Result<Vec<Vec<AggregationBucket>>> {
+    if buckets.is_empty() {
```
yhmo: src/v2/types/dql.rs line:2666
High ---- The zero-bucket early return drops the per-query shape: a normal aggregation response with nq=2 and agg_topks=[0,0] becomes an empty outer vector instead of two empty inner vectors, and a zero-bucket response with missing or mismatched agg_topks bypasses validation. The loop also infers the group count from topks.len() without requiring it to equal num_queries, so [3] for nq=2 is accepted as one group. Please treat both buckets and topks as empty only as no aggregation; otherwise validate topks arity and construct one possibly empty group per query, with zero-bucket and arity tests. This follows up the outdated grouping thread with current-head edge cases.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,915 @@
...
+        if let Some(tie_break_col) = tie_break_col {
+            inputs.push(tie_break_col.clone());
+            op = op.param("tie_break_col", tie_break_col);
+        }
+        self.ops.push(op.inputs(inputs));
+        self
+    }
+
+    /// Appends a `limit` operation with an optional offset.
+    pub fn limit(mut self, limit: i64, offset: i64) -> Self {
```
yhmo, line 625: src/v2/types/function_chain.rs line:486
Medium ---- FunctionChainOp is public, but FunctionChain exposes no add_op/op equivalent, so downstream callers cannot attach a raw or future operation such as filter or merge even though the pinned proto and the C++ 3.0 API provide FunctionChain::AddOp. That leaves the generic op type constructible but unusable and makes the module doc’s ordered FunctionChainOp composition incomplete. Please add a consuming add_op(FunctionChainOp) helper and cover it in validation tests while retaining the map/sort/limit conveniences.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,915 @@
...
+///////////////////////////////////////////////////////////////////////////////
+// fn_
+///////////////////////////////////////////////////////////////////////////////
+/// Convenience constructors for common server-side function chain expressions.
+pub mod fn_ {
+    use super::{FunctionChainArg, FunctionChainExpr};
+    use std::collections::HashMap;
+
+    /// Builds a numeric combination expression over two or more columns.
+    pub fn num_combine(
```
yhmo, line 982: src/v2/types/function_chain.rs line:628
Medium ---- Reopening the resolved helper-validation thread with current-head evidence: num_combine still accepts zero or one column, weighted mode without weights, mismatched weights, and literal args despite documenting two or more columns; decay, round_decimal, xgboost, and rerank_model likewise omit the PyMilvus function/range/column/non-empty-query checks. These type-correct values pass FunctionChain::validate and produce delayed server errors. Please enforce the known-expression contracts during validation or make the helpers return Result, with representative invalid-input tests.

src/v2/types/dql.rs
```diff
@@ -2347,6 +2347,371 @@ fn result_json_kind(value: &Value) -> &'static str {
...
+        let end = offset.checked_add(size).ok_or_else(|| {
+            Error::MalformedResponse("aggregation agg_topks overflowed usize".into())
+        })?;
+        if end > parsed.len() {
+            return Err(Error::MalformedResponse(format!(
+                "aggregation bucket count mismatch: agg_topks sum exceeds {} parsed buckets",
+                parsed.len()
+            )));
+        }
+        groups.push(parsed[offset..end].to_vec());
```
yhmo: src/v2/types/dql.rs line:2703
Medium ---- Grouping deep-clones every already-owned AggregationBucket tree through slice::to_vec(). Aggregation responses scale with nq times top-level buckets times nested groups, hits, and byte fields, so this duplicates the entire decoded payload after protobuf conversion. Please move the owned buckets into groups, for example by validating topks first and then splitting or decoding directly per query, and retain a multi-query grouping regression test.

src/v2/types/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/function_chain.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/function_chain.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,1270 @@
...
+                        ));
+                    }
+                };
+                if !matches!(function, "gauss" | "exp" | "linear") {
+                    return Err(Error::validation(
+                        "function_chains".into(),
+                        format!("unsupported decay function {function}"),
+                    ));
+                }
+                for name in ["origin", "scale", "offset", "decay"] {
```
yhmo, line 823: src/v2/types/function_chain.rs line:685
Low ---- decay only validates finiteness/presence of origin/scale/offset/decay; it does not enforce the server numeric ranges (the Milvus DecayExpr rejects scale <= 0, offset < 0, and decay outside (0, 1), also requiring [0.001, 0.999] for numerical stability). Type-correct input such as fn_::decay(col("ts"), "exp", 1700000000.0, -1.0, None, None) or a decay of 2.0 passes FunctionChain::validate and only fails later as a server-side error. Since this validation already enforces round_decimal [0, 6] and num_combine modes/weights, applying the same range checks here (scale > 0, offset >= 0, 0 < decay < 1) would make the failure local and consistent. Non-blocking; pymilvus and milvus-sdk-java also omit these range checks today.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,1270 @@
...
+// limitations under the License.
+
+//! Function-chain types for search-time reranking and post-processing.
+//!
+//! Function chains compose ordered [`FunctionChainOp`] operations (for example `map`, `sort`,
+//! `limit`) into a [`FunctionChain`] executed by the server at a given [`FunctionChainStage`].
+//! Expressions are built with [`FunctionChainExpr`] using [`col`] references and typed
+//! [`FunctionParamValue`] parameters. The [`fn_`] module provides convenience constructors for
+//! the common server-side functions, mirroring pymilvus's `fn` module.
+//!
```
yhmo, line 24: src/v2/types/function_chain.rs line:24
Low ---- Both new features ship with extensive unit tests and a compiling doctest but no runnable examples/v2 example, while milvus-sdk-java ships FunctionChainExample and SearchAggregationExample for the same features. An examples/v2 example (e.g. following the rerank_function.rs pattern) would make the end-to-end chain/aggregation flow discoverable and double as a system-test seed against a live server. Non-blocking.

src/v2/types/function_chain.rs
yhmo: Acknowledged. Keeping the current finiteness and presence checks without adding the server range validation: the Milvus server rejects invalid decay ranges at execution time, and both pymilvus and milvus-sdk-java omit these range checks today, so enforcing them here would make the Rust SDK stricter than the peer SDKs without a correctness gain.

src/v2/types/function_chain.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/function_chain.rs
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

src/v2/request/dql.rs
```diff
@@ -1871,6 +1917,24 @@ fn validate_search_request(value: &SearchRequest) -> Result<()> {
     validate_search_extra_params(&value.extra_params)?;
     validate_finite_range_parameter("radius", value.radius)?;
     validate_finite_range_parameter("range_filter", value.range_filter)?;
+    if !value.function_chains.is_empty() && value.rerank.is_some() {
```
yhmo, line 1921: src/v2/request/dql.rs line:1920
Low ---- The two new search-request mutual-exclusion checks here (`function_chains` + `rerank`, and `search_aggregation` + `group_by_field`) are user-visible validation behavior, but neither is exercised by a direct unit test in `src/v2/request/dql.rs` or `tests/v2/ut/dql.rs` today. The chain-level and aggregation-level `validate()` methods are well covered in `src/v2/types/function_chain.rs` and `src/v2/types/aggregation.rs`, and the iterator rejection path has a test in `src/v2/client/iterator.rs`, but a `SearchRequestBuilder` test that builds each illegal combination and asserts the `Validation` error (with the right `parameter` field) would lock in this contract. Suggest adding one test that asserts both rejections and one that confirms `function_chains`/`search_aggregation` survive `into_proto` encoding on the happy path.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,1270 @@
...
+    }
+
+    /// Returns the ordered operations.
+    pub fn get_ops(&self) -> &[FunctionChainOp] {
+        &self.ops
+    }
+
+    /// Validates that the chain can be attached to a search request.
+    pub fn validate(&self) -> Result<()> {
+        if self.stage == FunctionChainStage::Unspecified {
```
yhmo, line 652: src/v2/types/function_chain.rs line:524
Low ---- validate() only rejects the Unspecified stage, but Milvus restricts search-time function chains to the L0/L1/L2 rerank stages: splitFunctionChainsByStage in internal/proxy/function_chain_validator.go returns "function chain[%d] stage %s is not supported in search request" for Ingestion/PreProcess/PostProcess. A chain built with .stage(FunctionChainStage::Ingestion) passes SearchRequestBuilder::build() and then fails only on the server, which contradicts the "Validates that the chain can be attached to a search request" doc on this method. Consider rejecting the search-unsupported stages here, or wording the doc/error to say only that the stage must be non-Unspecified while the server enforces the search-stage subset.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,1270 @@
...
+                }
+                "sort" => {
+                    if let Some(FunctionParamValue::String(column)) = op.params.get("column") {
+                        if column.is_empty() {
+                            return Err(Error::validation(
+                                "function_chains".into(),
+                                "sort operation must not use an empty column name".into(),
+                            ));
+                        }
+                    } else {
```
yhmo: src/v2/types/function_chain.rs line:564
Low ---- The sort branch requires a "column" param, but the server derives the sort column from inputs[0] (NewSortOpFromRepr: column := repr.Inputs[0]) and never reads a "column" param. schema.proto documents sort as an input-driven op ("sort.inputs = ["$score"]"), so a raw FunctionChainOp::new("sort").inputs(["$score"]).param("desc", true) -- a valid, server-accepted payload -- is rejected here with "sort operation must specify a column". The fluent sort() helper writes both inputs and the redundant param, so only add_op() users hit this. Consider validating against inputs (matching the server contract) instead of the param, or documenting that the param is required.

src/v2/request/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/function_chain.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/function_chain.rs
yhmo: no need

src/v2/types/dql.rs
```diff
@@ -2347,6 +2347,628 @@ fn result_json_kind(value: &Value) -> &'static str {
...
+
+    /// Returns the field value.
+    pub fn get_value(&self) -> &AggregationBucketValue {
+        &self.value
+    }
+
+    pub(crate) fn from_proto(value: schema::BucketKeyEntry) -> Result<Self> {
+        Ok(Self {
+            field_id: value.field_id,
+            field_name: value.field_name,
```
yhmo: src/v2/types/dql.rs line:2722
Medium ---- The current decoder copies field_name verbatim, so a protobuf response with an empty name exposes BucketKeyEntry::get_field_name() == "" (and AggregationHitField has the same issue). C++/Java/PyMilvus substitute the numeric field ID when the name is absent; please mirror that fallback or reject the empty name as MalformedResponse, with decode tests for both key and hit fields.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,1430 @@
...
+                if !matches!(
+                    mode,
+                    "multiply" | "sum" | "max" | "min" | "avg" | "weighted"
+                ) {
+                    return Err(Error::validation(
+                        "function_chains".into(),
+                        format!("unsupported num_combine mode {mode}"),
+                    ));
+                }
+                match expr.get_params().get("weights") {
```
yhmo, line 783: src/v2/types/function_chain.rs line:775
Medium ---- Follow-up to the resolved generic-validation thread: this weighted branch checks only array length. A raw expression can still put String/Bool elements (or NaN/infinite Doubles) in weights and pass build, and empty parameter keys also serialize in expression/op maps. Please recursively reject empty map keys and require finite numeric weight elements, with hostile raw-value tests.

src/v2/types/dql.rs
```diff
@@ -2427,6 +3051,19 @@ impl SearchResults {
         self.recalls.push(value);
         self
     }
+
+    /// Sets the hierarchical aggregation buckets, grouped per query vector, and returns the
+    /// updated value.
+    pub fn agg_buckets(mut self, value: Vec<Vec<AggregationBucket>>) -> Self {
```
yhmo, line 3065: src/v2/types/dql.rs line:3057
Low ---- Follow-up to the resolved setter-family thread: the grouped rewrite adds agg_buckets only as a consuming setter, unlike adjacent results/recalls which provide set_results/set_recalls for in-place mutation. Please add set_agg_buckets(&mut self, ...) so the new member follows the same reusable-type API family; non-blocking consistency suggestion.

src/v2/response/dql.rs
```diff
@@ -1085,6 +1085,11 @@ impl SearchResponse {
             results: SearchResults {
                 results: single_results,
                 recalls: result.recalls,
+                agg_buckets: group_aggregation_buckets(
```
yhmo, line 1088: src/v2/response/dql.rs line:1088
Low ---- The grouping tests exercise group_aggregation_buckets directly, but no test sends a realistic aggregate SearchResults through SearchResponse::from_proto with num_queries/topks zero rows plus agg_buckets/agg_topks. Please add that end-to-end decoder test (for example nq=2, topks=[0,0], agg_topks=[1,0]) so the public response path, not only the helper, is locked in.

src/v2/types/function_chain.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/response/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/response/dql.rs
```diff
@@ -23,7 +23,10 @@
 use crate::proto::{milvus, schema};
 use crate::v2::error::{Error, Result};
 use crate::v2::types::Ids;
-use crate::v2::types::{DataType, FieldData, SparseVector};
+use crate::v2::types::{
+    group_aggregation_buckets, AggregationBucket, AggregationBucketValue, DataType, FieldData,
```
yhmo: src/v2/response/dql.rs line:27
Medium ---- The two aggregation imports added here are unused in this file: `AggregationBucket` is never referenced anywhere, and `AggregationBucketValue` is only referenced by the test module, which imports its own copy at line 1751. The CI doc step runs `cargo doc --no-deps` with `RUSTDOCFLAGS: -D warnings`, so these surface as `unused_imports` errors there. Please keep only `group_aggregation_buckets` in this import and rely on the test module's local import.

src/v2/request/dql.rs
```diff
@@ -1871,6 +1917,24 @@ fn validate_search_request(value: &SearchRequest) -> Result<()> {
...
+        return Err(Error::validation(
+            "function_chains".into(),
+            "cannot be used together with rerank".into(),
+        ));
+    }
+    for chain in &value.function_chains {
+        chain.validate()?;
+    }
+    if let Some(aggregation) = &value.search_aggregation {
```
yhmo, line 1930: src/v2/request/dql.rs line:1929
Medium ---- This aggregation validation block enforces mutual exclusion with `group_by_field`, but a non-zero search `offset` combined with `search_aggregation` still builds successfully and is only rejected by the server ("offset is not supported with search_aggregation", internal/proxy/task_search.go). The `SearchAggregation`/`search_aggregation` docs warn that `limit` is ignored but never mention that `offset` is rejected, so this is a trap for callers. Please reject `offset > 0` here (with a builder test) to match the server contract and this PR's goal of failing invalid requests locally.

yhmo, line 1930: src/v2/request/dql.rs line:1929
Low ---- `search_aggregation` combined with `highlighter` also passes `build()` today and is rejected only by the server ("highlighter and search_aggregation cannot be used simultaneously", internal/proxy/task_search.go). Consider rejecting that combination here too, for parity with the aggregation/`group_by_field` check just above.

src/v2/types/function_chain.rs
```diff
@@ -0,0 +1,1551 @@
...
+        self.stage
+    }
+
+    /// Returns the ordered operations.
+    pub fn get_ops(&self) -> &[FunctionChainOp] {
+        &self.ops
+    }
+
+    /// Validates that the chain can be attached to a search request.
+    pub fn validate(&self) -> Result<()> {
```
yhmo, line 651: src/v2/types/function_chain.rs line:651
Low ---- `validate()` does not require the chain to contain at least one operation. A chain built as `FunctionChain::new().stage(FunctionChainStage::L2Rerank)` with no ops passes validation and is rejected only by the server ("function chain[i] must contain at least one op", newFunctionChainRerankMeta in internal/proxy/function_chain_validator.go). Since `validate()` already rejects the empty stage and empty op names, rejecting an empty `ops` list here would make the failure local and consistent.

src/v2/response/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/function_chain.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #150, opened 2026-08-28
Support resumable query iterator cursors

yhmo:


src/v2/client/iterator.rs
```diff
@@ -188,6 +207,7 @@ impl QueryIterator {
                 "reduce_stop_for_best",
                 "false".into(),
             );
+            self.apply_element_cursor_params();
```
yhmo: src/v2/client/iterator.rs line:210
Medium ---- In `seek_to_offset`, `apply_element_cursor_params()` mutates `self.request`, but the RPC at the end of the loop sends the local `request` clone that was built *before* this call, so seek requests never carry `query_iter_last_pk`/`query_iter_last_element_offset`. For `element_filter` iterators with `offset > 0`, the server then re-serves the last primary key starting from element 0, inflating the seeked row count and making the offset overshoot (rows are skipped or duplicated in the final result). `next()` applies the same call to `self.request` before cloning it into the RPC, so normal pagination is unaffected; pymilvus's `__seek_to_offset` explicitly adds these params to the sent seek request (pymilvus/client/iterator/query_iterator.py). Suggest applying the call to the local `request` instead, and adding a test that resumes an element-filter cursor with a non-zero offset.

src/v2/response/dql.rs
```diff
@@ -790,6 +790,16 @@ impl QueryResponse {
     }
 
     pub(crate) fn from_proto(value: milvus::QueryResults) -> Result<Self> {
+        let element_indices = value
```
yhmo, line 799: src/v2/response/dql.rs line:793
Medium ---- `from_proto` decodes `element_indices` without validating that its length matches the output row count. pymilvus raises on `len(element_indices) != num_entities`, and the search decoder in this same file rejects short element indices with `MalformedResponse`. If the server returns fewer indices than rows, the mismatch is silently accepted: `split_at` can misalign the per-page indices and `query_response_cursor` may read the wrong (or no) element offset, corrupting a resumed element-filter cursor. Decode as `MalformedResponse` when `element_indices` is non-empty but its length differs from the row count.

src/v2/types/dql.rs
```diff
@@ -2977,6 +3005,84 @@ pub(crate) fn group_aggregation_buckets(
...
+            session_ts,
+            pk: QueryCursorPk::VarChar(pk.into()),
+            last_element_offset: None,
+        }
+    }
+
+    /// Sets the element position within the last primary key and returns the updated value.
+    ///
+    /// Only meaningful for `element_filter` iterators; ordinary iterators leave it `None`.
+    pub fn last_element_offset(mut self, value: i64) -> Self {
```
yhmo: src/v2/types/dql.rs line:3064
Low ---- `last_element_offset` accepts any `i64`, so a caller can build a cursor with a negative element offset. That value is then serialized verbatim as `query_iter_last_element_offset` by `apply_element_cursor_params` and sent to the server. Validate that the offset is non-negative (in the builder or in `query_iterator_with_cluster`) so the misuse fails locally with a clear validation error instead of reaching the server.

src/v2/types/dql.rs
```diff
@@ -1297,6 +1299,32 @@ impl QueryResults {
         self
     }
 
+    /// Sets the per-row element indices and returns the updated value.
+    ///
+    /// Each inner list holds the element offsets within an entity when an `element_filter`
+    /// iterator is used; ordinary queries leave this empty.
+    pub fn element_indices(mut self, value: Vec<Vec<i64>>) -> Self {
```
yhmo, line 1306: src/v2/types/dql.rs line:1306
Low ---- The Rust `QueryResults` keeps one row per entity and never expands `element_indices` into per-element rows, while pymilvus expands each entity into one row per matched element (see pymilvus grpc_handler.py element_indices expansion). For an `element_filter` iterator this means `batch_size`, `limit`, and row counts (and therefore cursor resume positions) are counted in entities, not matched elements, which diverges from pymilvus. Is entity-level granularity the intended Rust contract for this cursor feature, or should rows be expanded like pymilvus?

src/v2/client/iterator.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/response/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/dql.rs
yhmo: Entity-level granularity is intentional for the Rust SDK: `QueryIterator::next()` returns one row per matched entity (the same row model as ordinary query results), and `element_indices` is used only to capture the precise `last_element_offset` so a resumed `element_filter` iterator continues within the last primary key. Expanding each entity into one row per matched element (as pymilvus does) would change the row model and is a separate feature, so I am keeping the current contract.

tests/v2/ut/iterator.rs
```diff
@@ -562,6 +562,240 @@ async fn query_iterator_advances_by_primary_key_beyond_query_window() {
...
+                .query(
+                    QueryRequest::builder()
+                        .collection_name("books")
+                        .filter("large_query_iterator")
+                        .output_fields(["id"])
+                        .build()
+                        .expect("valid request"),
+                )
+                .batch_size(8_192)
+                .cursor(QueryCursor::int64(0, 8_191))
```
yhmo, line 581: tests/v2/ut/iterator.rs line:581
Medium ---- The new resumable-cursor feature is only exercised with Int64 primary keys; the VarChar path (`QueryCursorPk::VarChar`, `QueryCursor::varchar`, the JSON-quoted cursor predicate in `next_filter`, and the raw-string `query_iter_last_pk` parameter in `apply_element_cursor_params`) has no test. The closest existing test (`legacy_varchar_primary_keys_are_escaped_in_exclusion_filters`) covers only the V1 search path. A VarChar resume test would guard the trickiest part of this change: the cursor value is serialized differently in the filter expression (JSON-quoted) versus the `query_iter_last_pk` query parameter (raw).

src/v2/types/dql.rs
```diff
@@ -2977,6 +3005,84 @@ pub(crate) fn group_aggregation_buckets(
...
+pub enum QueryCursorPk {
+    /// Int64 primary-key position.
+    Int64(i64),
+    /// VarChar primary-key position.
+    VarChar(String),
+}
+
+impl QueryCursor {
+    /// Creates a cursor with an Int64 primary-key position.
+    pub fn int64(session_ts: u64, pk: i64) -> Self {
```
yhmo, line 3072: src/v2/types/dql.rs line:3044
Low ---- `session_ts` is a required public constructor argument, but its semantics are undocumented: a value of `0` makes the SDK run the server-side timestamp probe and derive the guarantee timestamp, while any non-zero value pins the resumed iterator to that MVCC snapshot and skips the probe (overriding the resumed query's own consistency-derived timestamp). Callers building a cursor manually have no way to know when to pass `0`. Consider documenting this in the rustdoc (and on `QueryCursor::varchar`) or providing a dedicated constructor for captured cursors.

tests/v2/ut/iterator.rs
yhmo: Fixed. Added a VarChar resume test `query_iterator_var_char_cursor_resumes_from_quoted_predicate_and_raw_last_pk` in tests/v2/ut/iterator.rs: it resumes an element-filter iterator over a VarChar-PK collection and asserts the resumed request (a) JSON-quotes the VarChar cursor in the filter expression (`title >= "row-0"`), (b) passes the raw string (not the JSON-quoted form) in the `query_iter_last_pk` query parameter, and (c) carries `query_iter_last_element_offset`. A matching `varchar_element_query_iterator` branch was added to the mock server in tests/v2/ut/common.rs. All `cargo test --test v2_ut` (99) and `cargo test --lib` (860) pass.

yhmo: Fixed. Added a VarChar resume test `query_iterator_var_char_cursor_resumes_from_quoted_predicate_and_raw_last_pk` in tests/v2/ut/iterator.rs: it resumes an element-filter iterator over a VarChar-PK collection and asserts the resumed request (a) JSON-quotes the VarChar cursor in the filter expression (`title >= "row-0"`), (b) passes the raw string (not the JSON-quoted form) in the `query_iter_last_pk` query parameter, and (c) carries `query_iter_last_element_offset`. A matching `varchar_element_query_iterator` branch was added to the mock server in tests/v2/ut/common.rs. All `cargo test --test v2_ut` (99) and `cargo test --lib` (860) pass.

yhmo: Fixed. Added a VarChar resume test `query_iterator_var_char_cursor_resumes_from_quoted_predicate_and_raw_last_pk` in tests/v2/ut/iterator.rs: it resumes an element-filter iterator over a VarChar-PK collection and asserts the resumed request (a) JSON-quotes the VarChar cursor in the filter expression (`title >= "row-0"`), (b) passes the raw string (not the JSON-quoted form) in the `query_iter_last_pk` query parameter, and (c) carries `query_iter_last_element_offset`. A matching `varchar_element_query_iterator` branch was added to the mock server in tests/v2/ut/common.rs. All `cargo test --test v2_ut` (99) and `cargo test --lib` (860) pass.

src/v2/types/dql.rs
yhmo: Fixed. Documented the `session_ts` semantics on both `QueryCursor::int64` and `QueryCursor::varchar` constructors in src/v2/types/dql.rs: passing `0` makes the SDK run the server-side timestamp probe and derive the guarantee timestamp (like a fresh iterator), while a non-zero captured timestamp pins the resumed iterator to that MVCC snapshot and skips the probe. Also pointed callers to the captured cursor from `QueryIterator::cursor` via `QueryCursor::session_ts`. `cargo doc` builds without new warnings.

src/v2/client/iterator.rs
```diff
@@ -174,6 +188,15 @@ impl QueryIterator {
         Ok(())
     }
 
+    /// Returns the primary-key cursor position after the last page read, if any.
+    ///
+    /// Pass the returned cursor to
+    /// [`QueryIteratorRequestBuilder::cursor`](crate::v2::request::dql::QueryIteratorRequestBuilder::cursor)
+    /// on a new iterator to resume pagination from exactly this position.
```
yhmo, line 203: src/v2/client/iterator.rs line:195
Medium ---- Resuming still runs `seek_to_offset(offset)` unconditionally (line 840), so a resumed iterator that retains a non-zero `query.offset` skips `offset` additional rows *past* the captured cursor. pymilvus's `QueryIterator.__seek_to_offset` returns early when a cursor is already present (`_next_id is not None`), so the reference SDK never re-applies offset on resume; reusing the original request here therefore double-counts the offset and silently skips rows the user already consumed, contradicting this doc's "resume pagination from exactly this position". If re-applying offset relative to the cursor is intentional (the new element-cursor seek test suggests it is), the docs should state it explicitly and warn that resuming with the original offset creates a gap.

src/v2/request/dql.rs
```diff
@@ -1763,6 +1770,16 @@ impl QueryIteratorRequestBuilder {
         self
     }
 
+    /// Resumes pagination from a previously captured primary-key cursor and returns the updated
+    /// value.
+    ///
+    /// The cursor type must match the collection primary key (`Int64` for Int64 keys,
+    /// `VarChar` for VarChar keys). The iterator starts immediately after the cursor value.
```
yhmo: src/v2/request/dql.rs line:1777
Low ---- "The iterator starts immediately after the cursor value" holds for plain cursors (the generated predicate is `pk > value`) but not for element-filter cursors: `next_filter` emits `pk >= value` and the server resumes at the cursor pk from the recorded element offset, so the resumed iterator starts *at* the cursor value rather than after it. Please clarify the wording so callers do not expect a strict-exclusive resume for element cursors.

src/v2/client/iterator.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/iterator.rs
```diff
@@ -188,6 +211,11 @@ impl QueryIterator {
                 "reduce_stop_for_best",
                 "false".into(),
             );
+            apply_element_cursor_params(
```
yhmo: src/v2/client/iterator.rs line:214
Medium ---- `seek_to_offset` now applies `query_iter_last_pk`/`query_iter_last_element_offset` to the seek request once an element cursor is established, but that request also sets `iterator=false`. The Milvus server rejects `query_iter_last_*` unless `iterator=true` (`internal/proxy/task_query.go`, `parseQueryIteratorCursor`), so an `element_filter` iterator with `offset > MAX_BATCH_SIZE` (16384) hard-fails with a parameter error on the second seek batch. The new test `query_iterator_element_cursor_seek_requests_carry_element_params` codifies this behavior against the mock server, which does not enforce the real server contract. Please verify against a real server and align the seek path (e.g., drop the element params on `iterator=false`, or document/guard the limitation).

src/v2/types/dql.rs
```diff
@@ -2977,6 +3005,96 @@ pub(crate) fn group_aggregation_buckets(
...
+    }
+
+    /// Returns the primary-key position after the last page read.
+    pub fn pk(&self) -> &QueryCursorPk {
+        &self.pk
+    }
+
+    /// Returns the element position within the last primary key, if this is an `element_filter`
+    /// cursor.
+    pub fn get_last_element_offset(&self) -> Option<i64> {
```
yhmo, line 3148: src/v2/types/dql.rs line:3093
Low ---- `get_last_element_offset()` is the only accessor in `QueryCursor` that uses the `get_` prefix; its siblings are `session_ts()` and `pk()`. The V2 accessor convention reserves `get_*` for keyed lookups and prefers plain field-named accessors, so `last_element_offset()` would be consistent here.

src/v2/client/iterator.rs
```diff
@@ -155,11 +164,16 @@ impl QueryIterator {
             self.finished = true;
             return Ok(None);
         }
-        let cursor =
-            query_response_cursor(&response, &self.primary_field_name, self.primary_field_type)?;
         let next_remaining = self.remaining.map(|left| left.saturating_sub(count));
         let next_finished = next_remaining == Some(0) || (source_exhausted && next_cache.is_none());
 
+        let cursor = query_response_cursor(
```
yhmo, line 178: src/v2/client/iterator.rs line:170
Medium ---- For `element_filter` iterators, page exhaustion is judged by entity count while the server limits element-level pages by element count, so iteration can stop early when `reduce_stop_for_best` is false.

`source_exhausted = count < batch_size` compares the number of returned *entities* to `batch_size`, but for an element-level query with `reduce_stop_for_best=false` the proxy applies `elementLimit = batch_size` (internal/proxy/task_query.go: `IReduceInOrder` -> `ShouldUseInputLimit` true -> `elementLimit`), so each page returns however many entities fit within `batch_size` *elements*. When a single entity matches several elements, a page returns fewer than `batch_size` entities while more matching entities remain: `source_exhausted` becomes true, `finished` is set, and rows are silently dropped without the caller seeing an error. The mock always emits one element per entity (`element_indices: data: vec![start + i]`), so the new tests never hit this path. Consider deriving `source_exhausted` from the summed `element_indices` lengths for element iterators, or documenting/guarding the `reduce_stop_for_best=false` limitation.

tests/v2/ut/common.rs
```diff
@@ -1634,18 +1634,77 @@ impl MilvusService for MockMilvus {
             } else if request.expr.contains("element_filter_query_iterator") {
                 let start = request
                     .expr
-                    .split_once("id > ")
+                    .split_once("id >=")
```
yhmo, line 1637: tests/v2/ut/common.rs line:1637
Low ---- The element-filter mock treats `id >= X` as exclusive (`start = X + 1`) and reports the primary-key value as the element offset (`data: vec![start + i]`), so the resume tests validate request construction but never exercise the server-side resume contract.

On a real server, resuming an element cursor re-enters the last primary key at `last_element_offset + 1`. Because the mock never emits a page that starts inside the cursor primary key and effectively treats `>=` as `>`, the `pk >=` + `query_iter_last_element_offset` path is verified only on the request side; the response side (`cursor.last_element_offset` captured from `get_element_indices().last()`) is not exercised, so a regression in element-offset resume could pass CI. Consider having the mock return the cursor primary key with the remaining element offsets so the resume semantics are pinned end to end.

src/v2/types/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/iterator.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/iterator.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

tests/v2/ut/common.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/iterator.rs
```diff
@@ -174,11 +196,20 @@ impl QueryIterator {
...
+    pub fn cursor(&self) -> Option<QueryCursor> {
+        self.cursor.clone()
+    }
+
     async fn seek_to_offset(&mut self, mut offset: usize) -> Result<()> {
         while offset > 0 {
             let size = offset.min(MAX_BATCH_SIZE);
             let mut request = self.request.clone();
-            request.expr = self.next_filter();
+            request.expr = self.next_filter_exclusive();
```
yhmo, line 212: src/v2/client/iterator.rs line:212
Medium ---- The offset seek for `element_filter` iterators counts by entity while the server caps each page by element count, so an element iterator with a non-zero offset silently truncates results. `seek_to_offset` sends `iterator=false`, `reduce_stop_for_best=false`, `limit=size`, and then marks the iterator finished when `query_row_count(&raw) < size` (lines 238-241). For an element-level query the server applies `elementLimit = limit` (internal/proxy/task_query.go: `IReduceNoOrder` -> `ShouldUseInputLimit` true -> `elementLimit`), so a seek page returns at most `size` *elements*; when an entity matches several elements the page carries fewer *entities* than `size`, `count < size` becomes true and the iterator stops even though more matching rows remain. This is the same entity-vs-element mismatch that `next()` already fixes with `query_response_element_count`, but the seek path still uses `query_row_count`. The new test `query_iterator_element_cursor_seek_requests_omit_element_params` cannot catch it because the `element_filter_query_iterator` mock emits exactly one element per entity. Suggest judging seek-page exhaustion by summed element counts for element iterators, or documenting/guarding the limitation.

AGENTS.md
```diff
@@ -0,0 +1,96 @@
+# AGENTS.md
```
yhmo, line 1: AGENTS.md line:1
Low ---- This 96-line AGENTS.md (repo guidance for AI coding agents) is unrelated to the stated purpose of this PR, resumable query iterator cursors. It is useful documentation, but bundling it into the feature PR widens the review surface and makes the feature harder to isolate; consider landing it as a separate docs-only PR, or at least calling it out in the PR description so reviewers know it is intentional.

src/v2/client/iterator.rs
```diff
@@ -1285,10 +1408,29 @@ fn query_response_row_count(response: &response::dql::QueryResponse) -> usize {
...
+fn query_response_element_count(response: &response::dql::QueryResponse) -> usize {
+    let indices = response.results().get_element_indices();
+    if indices.is_empty() {
+        query_response_row_count(response)
+    } else {
+        indices.iter().map(Vec::len).sum()
+    }
+}
+
+fn is_element_filter_expr(expr: &str) -> bool {
```
yhmo, line 1449: src/v2/client/iterator.rs line:1424
Low ---- `is_element_filter_expr` detects element-filter iterators with an unanchored substring search (`expr.to_ascii_lowercase().contains("element_filter")`). Any plain filter that merely contains the text "element_filter" - e.g. a string literal `name == "element_filter"` or a JSON path `tags["element_filter"] == 1` - is misclassified as an element-filter iterator. The practical impact is currently muted only because a non-element query returns empty `element_indices`, so `query_response_element_count` falls back to the entity row count and `query_response_cursor` captures no element offset; but the flag silently changes the exhaustion path and the cursor/cache logic, making the classification fragile and hard to reason about. Consider detecting the actual `element_filter(` function call instead of an unanchored substring so a field name or string value cannot flip a plain iterator into element mode.

src/v2/types/dql.rs
```diff
@@ -2977,6 +3005,142 @@ pub(crate) fn group_aggregation_buckets(
...
+    VarChar(String),
+}
+
+impl QueryCursor {
+    /// Creates a value initialized with its SDK defaults.
+    ///
+    /// The default cursor carries an `Int64` primary-key position of `0` and no element offset.
+    /// Use [`Self::int64`]/[`Self::varchar`] for the primary construction path, or configure this
+    /// default with the fluent and in-place setters.
+    pub fn new() -> Self {
```
yhmo, line 3057: src/v2/types/dql.rs line:3048
Low ---- `QueryCursor::new()` produces a default cursor with `pk = Int64(0)` and `session_ts = 0`. The cursor type check in `query_iterator_with_cluster` only verifies the pk variant matches the collection primary key, so passing a freshly-constructed `QueryCursor::new()` to a query iterator on an Int64-PK collection is accepted and silently resumes from `pk > 0`, dropping every row with pk <= 0 (negative Int64 primary keys are valid). The rustdoc points users at `int64`/`varchar` for construction, but nothing in the type system or the `.cursor()` builder prevents this silent data loss. Consider making the default pk position an explicit "no position yet" state that resume rejects, or at minimum documenting/validating that a default cursor is never a valid resume point.

tests/v2/ut/iterator.rs
```diff
@@ -562,6 +562,503 @@ async fn query_iterator_advances_by_primary_key_beyond_query_window() {
...
+            .expect("captured resumed cursor")
+            .get_last_element_offset(),
+        Some(3)
+    );
+
+    server.shutdown().await;
+}
+
+#[tokio::test]
+async fn query_iterator_var_char_cursor_resumes_from_quoted_predicate_and_raw_last_pk() {
```
yhmo, line 931: tests/v2/ut/iterator.rs line:896
Medium ---- The resumable-cursor feature is only exercised for VarChar through the `element_filter` path (this test plus the `varchar_element_query_iterator` / `int64_reenter_element_query_iterator` mocks). The plain (non-element-filter) VarChar cursor path - `QueryCursor::varchar(...)`, the exclusive `pk > "value"` JSON-quoted predicate built by `build_filter(false)`, and resuming without any `query_iter_last_*` param - has no test, and the `QueryCursor::varchar` constructor is never invoked anywhere in the suite. That exclusive VarChar resume is precisely where the cursor value is serialized differently in the filter expression (JSON-quoted) versus the raw string form, so a regression there would go undetected. Consider a test that resumes a plain VarChar iterator from `QueryCursor::varchar` and asserts the `title > "value"` predicate and the absence of `query_iter_last_pk`.

src/v2/client/iterator.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/types/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/iterator.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

tests/v2/ut/iterator.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

AGENTS.md
yhmo: as designed

src/v2/client/iterator.rs
```diff
@@ -1281,14 +1412,50 @@ fn query_row_count(response: &milvus::QueryResults) -> usize {
...
+    expr.to_ascii_lowercase().contains("element_filter(")
```
yhmo: src/v2/client/iterator.rs line:1450
Medium ---- is_element_filter_expr requires the exact substring "element_filter(", but Milvus's Plan.g4 skips whitespace between tokens (Whitespace: [ \t]+ -> skip), so a valid expression written "element_filter (tags, tag == \"x\")" (space before the paren) is classified as a plain iterator. For multi-element entities this silently falls back to entity-count exhaustion (count < batch_size), reintroducing the premature-termination/row-loss this PR fixes, and the query_iter_last_* resume params are never emitted. pymilvus detects with a bare-token match ("element_filter" in expr.lower()), which handles the whitespace variant. Consider matching the function token loosely (e.g. "element_filter" optionally followed by whitespace then "(") or a small regex, and add a regression test.

yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #151, opened 2026-09-01
Support nullable struct field

yhmo:


src/v2/types/collection.rs
```diff
@@ -1040,7 +1044,11 @@ impl StructFieldSchema {
                     field.data_type = field.element_type;
                     field.element_type = schema::DataType::None as i32;
                     field.type_params.retain(|pair| pair.key != "max_capacity");
-                    FieldSchema::from_proto(field)
+                    let mut field = FieldSchema::from_proto(field)?;
+                    // A struct's nullable is applied to every sub-field on the wire, so the
+                    // decoded sub-fields inherit the struct's nullable.
+                    field.nullable = value.nullable;
```
yhmo, line 1050: src/v2/types/collection.rs line:1050
Medium ---- This change makes from_proto() give every decoded sub-field of a nullable struct `nullable=true`, but the validate() rule in this same PR (line 950) now rejects any nullable sub-field with "struct sub-fields cannot be nullable individually". As a result, a schema decoded from describe_collection (nullable struct -> sub-fields nullable=true) can never pass StructFieldSchema::validate() again: describe -> recreate (CreateCollectionRequest) and AddCollectionStructFieldRequest::build both re-validate and will fail. The system test only asserts decoded sub-fields are nullable without re-encoding them, so this round-trip break is not covered. Either let from_proto preserve non-nullable sub-fields under a nullable parent (and drop the wire propagation), or relax validate() to allow sub-fields matching the parent nullable, otherwise describe/recreate of a nullable-struct collection breaks.

src/v2/types/collection.rs
```diff
@@ -947,10 +947,10 @@ impl StructFieldSchema {
                     "struct sub-fields cannot be clustering keys".into(),
                 ));
             }
-            if !self.nullable && field.nullable {
+            if field.nullable {
```
yhmo: src/v2/types/collection.rs line:950
Medium ---- StructFieldSchema::validate() now rejects any nullable sub-field, but from_proto() (line ~1050) sets every decoded sub-field's nullable to the struct's nullable. A nullable struct decoded via describe_collection therefore carries nullable sub-fields and can no longer pass validate(); reusing that schema in create_collection or add_collection_struct_field fails with "struct sub-fields cannot be nullable individually". This breaks the describe->create round-trip for nullable structs (the ST test asserts sub-fields inherit nullable but never re-validates). Consider allowing a sub-field nullable that equals the parent struct's nullable, or stop propagating nullable to sub-fields in from_proto() and align the round-trip test accordingly.

src/v2/response/dql.rs
```diff
@@ -573,59 +573,93 @@ fn struct_field_data(name: String, value: schema::StructArrayField) -> Option<Fi
...
-                            is_dynamic: false,
-                            valid_data: Vec::new(),
-                            field: Some(proto_field_data::Field::Vectors(vectors)),
-                            ..Default::default()
-                        })?)
-                    })
-                    .collect::<Option<Vec<_>>>()?
+                let full_length = parent_valid_data
+                    .as_ref()
+                    .is_none_or(|valid_data| array.data.len() == valid_data.len());
```
yhmo: src/v2/response/dql.rs line:624
Low ---- The full-length vs compacted decode branches are duplicated for scalar (Scalars, ~576-616) and vector (Vectors, ~622-662) sub-fields; only the decode_field_data call differs. Extracting a shared helper that takes a row-decoding closure would avoid the two paths drifting (e.g. one fixing placeholder alignment while the other is missed). This is a distinct concern from the existing thread about missing vector-path unit-test coverage.

src/v2/response/dql.rs
```diff
@@ -573,59 +573,93 @@ fn struct_field_data(name: String, value: schema::StructArrayField) -> Option<Fi
...
+                                field_id: field.field_id,
+                                is_dynamic: false,
+                                valid_data: Vec::new(),
+                                field: Some(proto_field_data::Field::Scalars(scalars)),
+                                ..Default::default()
+                            })?)
+                        })
+                        .collect::<Option<Vec<_>>>()?
+                };
+                align_struct_subfield(grouped, parent_valid_data.as_deref())?
```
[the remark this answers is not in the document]
yhmo: Added the companion test `query_decodes_compacted_nullable_struct_vector_subfield`: it builds a compacted nullable struct payload for a vector sub-field (`VectorArray`, same `[false, true, false]` validity mask) and verifies the decoder re-expands it back to the three logical rows with null rows decoding as null. Both scalar and vector compacted paths now have unit coverage.

src/v2/types/collection.rs
yhmo: Fixed: `StructFieldSchema::validate()` now only rejects a nullable sub-field when the parent struct is non-nullable (`field.nullable && !self.nullable`). A nullable struct's sub-fields inherit `nullable=true` on the wire and on decode, so a schema returned by describe_collection passes validate() again and describe -> recreate (CreateCollectionRequest / AddCollectionStructFieldRequest::build) works. Updated `struct_field_schema_allows_nullable_subfields_under_nullable_struct` and added a decode-then-revalidate assertion in the wire-propagation test.

src/v2/types/collection.rs
yhmo: Same root cause as the from_proto comment above; fixed together by relaxing validate() to allow a sub-field nullable that matches the parent struct's nullable. A decoded nullable struct (sub-fields nullable=true) now re-validates, so describe -> recreate works. Regression covered by `struct_field_schema_allows_nullable_subfields_under_nullable_struct` plus the decode-then-validate round-trip assertion.

src/v2/response/dql.rs
yhmo: Done: extracted a shared helper `decode_struct_subfield_rows(rows, valid_data, parent_valid_at, decode_row)` that handles the full-length vs compacted branching and placeholder alignment once, with the scalar/vector branches passing only their per-row decoding closure. The two decode paths can no longer drift on compaction handling. Unit tests cover both scalar and vector compacted payloads.

PR #152, opened 2026-09-01
Align search round_decimal validation range with pymilvus

yhmo:


src/v2/request/dql.rs
```diff
@@ -1933,10 +1933,10 @@ fn validate_search_request(value: &SearchRequest) -> Result<()> {
     positive_i64("limit", value.limit)?;
     non_negative_i64("offset", value.offset)?;
     positive_i64("group_size", value.group_size)?;
-    if value.round_decimal < -1 {
+    if !(-1..=6).contains(&value.round_decimal) {
```
yhmo, line 1936: src/v2/request/dql.rs line:1936
Low ---- round_decimal can still bypass this new client-side range check when supplied through extra_params (a String-valued map), because validate_search_extra_params only reserves "params", "topk", "anns_field", "metric_type" and into_proto forwards extra_params straight into the server search params. The server rejects out-of-range values anyway (proxy search_util.go enforces -1 or [0,6]), so this is a consistency gap rather than a correctness hole. Consider adding "round_decimal" to the RESERVED keys (and mirroring it for HybridSearchRequest rank params) so the guard cannot be circumvented.

yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #153, opened 2026-09-02
Align list_indexes field filter and empty-input DML with pymilvus

yhmo:


src/v2/client/index.rs
```diff
@@ -70,14 +70,23 @@ impl ClientV2 {
...
         &self,
         request: request::index::ListIndexesRequest,
     ) -> Result<response::index::ListIndexesResponse> {
+        let field_name = request.field_name().to_owned();
         let database = self.current_database();
         let response = rpc_with_retry!(self, get_index_statistics, request.into_proto(&database))?;
         status_to_result(&response.status)?;
-        response::index::ListIndexesResponse::from_proto(response)
+        let mut result = response::index::ListIndexesResponse::from_proto(response)?;
+        if !field_name.is_empty() {
```
yhmo: Because `GetIndexStatisticsRequest` in the pinned proto has no `field_name` field, this filter is applied purely client-side: the SDK always fetches and decodes every index description of the collection and then discards non-matching entries. Behavior is correct, but a caller passing `field_name` gets no server-side pruning, so all index metadata is still transferred. Consider noting in the rustdoc that the filter is applied locally after `get_index_statistics` (the Java SDK narrows server-side via `DescribeIndexRequest.field_name`).

tests/v2/ut/index.rs
```diff
@@ -214,3 +214,91 @@ async fn index_wait_timeout_bounds_sleep_and_stalled_poll() {
...
+                .build()
+                .expect("valid request"),
+        )
+        .await
+        .unwrap();
+    let mut all_names = all.index_names().to_vec();
+    all_names.sort();
+    assert_eq!(all_names, ["title_idx", "vector_idx"]);
+
+    let vector_only = client
```
yhmo, line 261: Coverage gap: no case combines `index_name` with `field_name` (server narrows by index name, then the client-side filter runs), and no case puts two indexes on the same field to prove `index_names` and `indexes` stay in sync after `filter_by_field` rebuilds `index_names`. A small added case for either would guard the combined-filter semantics.

src/v2/response/index.rs
```diff
@@ -157,6 +157,16 @@ impl ListIndexesResponse {
             indexes,
         })
     }
+
+    pub(crate) fn filter_by_field(&mut self, field_name: &str) {
```
yhmo: src/v2/response/index.rs line:161
Low ---- filter_by_field runs only after ListIndexesResponse::from_proto has already decoded every index description, so a malformed "params" JSON on an unrelated field aborts a field-scoped list_indexes call (from_proto propagates the serde_json error). The C++ SDK filters the raw proto index_descriptions by field_name before constructing IndexDesc, so unrelated malformed data cannot break a field-scoped listing. Consider filtering index_descriptions by field_name before decoding (or at least documenting this coupling), so a field-scoped query does not inherit decode failures from other fields.

src/v2/client/index.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

tests/v2/ut/index.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/response/index.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/client/index.rs
```diff
@@ -70,14 +70,23 @@ impl ClientV2 {
     }
 
     /// Lists index names and metadata associated with a collection.
+    ///
+    /// When the request carries a non-empty
+    /// [`request::index::ListIndexesRequest::field_name`], only indexes built on
+    /// that field are returned. The filter is applied locally after the
```
yhmo, line 76: src/v2/client/index.rs line:76
Low ---- On a real server, get_index_statistics returns an IndexNotExist error status when the collection has no index at all, so list_indexes fails instead of returning the empty result the new docstring implies for a field with no indexes. Both the C++ SDK (MilvusClientV2Impl::listIndexes) and Java SDK (IndexService.listIndexes) special-case IndexNotExist / code 700 into an empty list, and the mock get_index_statistics always returns success with empty descriptions, so the unit tests cannot observe this divergence. Consider treating IndexNotExist as an empty listing to match the other SDKs, or document the divergence.

yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/response/dml.rs
```diff
@@ -38,12 +38,9 @@ pub struct DmlResponse {
...
 impl DmlResponse {
-    ///////////////////////////////////////////////////////////////////////////////
-    // Test Cases
-    ///////////////////////////////////////////////////////////////////////////////
-
-    #[cfg(test)]
-    fn empty() -> Self {
+    /// Creates an empty response used when the client short-circuits an empty
+    /// mutation before issuing the RPC (matching pymilvus's empty-input behavior).
+    pub(crate) fn empty() -> Self {
```
yhmo, line 43: src/v2/response/dml.rs line:43
Low ---- DmlResponse::empty() hardcodes ids to Ids::default(), which always resolves to the Ids::Int64(vec![]) variant. On a collection whose primary key is VarChar, a real insert/upsert response carries Ids::VarChar, so a caller that pattern-matches the Ids variant (for example matches!(resp.ids(), Ids::VarChar(_))) will take the wrong branch for a short-circuited empty mutation. Consider deriving the ids variant from the resolved collection schema, or documenting that the ids variant of an empty short-circuit response is unspecified.

src/v2/client/dml.rs
```diff
@@ -33,6 +33,9 @@ impl ClientV2 {
         &self,
         request: request::dml::InsertRequest,
     ) -> Result<response::dml::InsertResponse> {
+        if request.columns.is_empty() && request.rows.is_empty() {
```
yhmo, line 36: src/v2/client/dml.rs line:36
Low ---- This empty-input short-circuit returns before telemetry.begin_operation, so empty insert/upsert (and empty-ids get in client/dql.rs) are invisible to the telemetry metrics that every other ClientV2 operation feeds. If telemetry is used to track operation counts, these no-op calls will undercount. Consider moving the short-circuit after begin_operation, or emitting a telemetry span for the empty path.

PR #154, opened 2026-09-02
Align compact, list_indexes, and round_decimal with pymilvus

yhmo:
## Summary

Align several V2 request parameters with the PyMilvus `MilvusClient` baseline for the 2.6 line.

### compact
- Reject `target_size <= 0` (previously `0` was accepted as "server default"; **behavior change on the 2.6 branch**: callers that passed `0` must now omit the field to let the server pick the default target size, matching pymilvus).
- Add `l0_compaction` parameter mapped to `ManualCompactionRequest.l0_compaction`.
- Add `target_size_unit` (`TargetSizeUnit`: B/KB/MB/GB/TB/PB) used to convert `target_size` to megabytes before it is sent on the wire.
- `target_size` and `target_size_unit` are normalized at `build()` time; a built request always reports the value in MB, so `into_builder().build()` round-trips idempotently.

### list_indexes
- Add `field_name` filter; returned indexes are filtered by field name when set (empty means all fields).

### search / hybrid_search / search_iterator
- Enforce `round_decimal` in `-1..=6`, matching pymilvus `is_legal_round_decimal`.

## Tests
- Unit tests for the new `compact` validation/unit conversion, `into_builder` idempotency, `list_indexes` field filter, and `round_decimal` upper-bound rejection.

### DIFF
--- a/src/v2/request/dql.rs
+++ b/src/v2/request/dql.rs
@@ -1597,10 +1597,10 @@ impl HybridSearchRequestBuilder {
         non_negative_i64("offset", self.value.offset)?;
         positive_i64("group_size", self.value.group_size)?;
         validate_search_extra_params(&self.value.extra_params)?;
-        if self.value.round_decimal < -1 {
+        if !(-1..=6).contains(&self.value.round_decimal) {
             return Err(Error::validation(
                 "round_decimal".into(),
-                "must be -1 or greater".into(),
+                "must be in -1..=6".into(),
             ));
         }
         Ok(self.value)
@@ -1843,10 +1843,10 @@ fn validate_search_request(value: &SearchRequest) -> Result<()> {
     positive_i64("limit", value.limit)?;
     non_negative_i64("offset", value.offset)?;
     positive_i64("group_size", value.group_size)?;
-    if value.round_decimal < -1 {
+    if !(-1..=6).contains(&value.round_decimal) {
         return Err(Error::validation(
             "round_decimal".into(),
-            "must be -1 or greater".into(),
+            "must be in -1..=6".into(),
         ));
     }
     validate_search_extra_params(&value.extra_params)?;
@@ -2228,6 +2228,34 @@ mod search_request_tests {
             result,
             Err(crate::v2::error::Error::Validation(error)) if error.parameter() == "ids"
         ));
+    }
+
+    #[test]
+    fn search_rejects_round_decimal_outside_of_legal_range() {
+        let search = || {
+            SearchRequest::builder()
+                .collection_name("books")
+                .vector_field("vector")
+                .vectors(SearchVectors::Float(vec![vec![0.1, 0.2]]))
+                .limit(3)
+        };
+        let hybrid = || {
+            let sub_request = SubSearchRequest::builder()
+                .vector_field("vector")
+                .vectors(SearchVectors::Float(vec![vec![0.1, 0.2]]))
+                .limit(3)
+                .build()
+                .expect("valid sub search request");
+            HybridSearchRequest::builder()
+                .collection_name("books")
+                .sub_requests(vec![sub_request])
+        };
+
+        assert!(search().round_decimal(7).build().is_err());
+        assert!(search().round_decimal(-2).build().is_err());
+        assert!(search().round_decimal(6).build().is_ok());
+        assert!(hybrid().round_decimal(7).build().is_err());
+        assert!(hybrid().round_decimal(6).build().is_ok());
     }
 
     #[test]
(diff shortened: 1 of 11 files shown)


src/v2/request/utility.rs
```diff
@@ -668,22 +686,53 @@ impl CompactRequestBuilder {
...
             &self.value.collection_name,
         )?;
-        if self.value.target_size < 0 {
-            return Err(Error::validation(
-                "target_size".into(),
-                "must not be negative".into(),
-            ));
-        }
+        self.value.target_size = match self.value.target_size {
+            Some(size) => Some(size_to_mb(size, self.value.target_size_unit)?),
```
yhmo: src/v2/request/utility.rs line:708
Medium ---- build() normalizes target_size to megabytes but keeps the original target_size_unit, and into_builder() preserves both, so rebuilding an already-built request silently double-converts: .target_size(2).target_size_unit(GB).build() stores Some(2048); request.into_builder().build() then runs size_to_mb(2048, GB) and sends 2097152 MB (~2 PB) to the server instead of 2048 MB. Consider converting only in into_proto() (keep the raw value + unit on the request) or resetting target_size_unit to MB after build, and add an into_builder() round-trip test.

src/v2/request/utility.rs
```diff
@@ -668,22 +686,53 @@ impl CompactRequestBuilder {
...
+            None => None,
+        };
         Ok(self.value)
     }
 }
 
+/// Converts a target size expressed in the given unit to megabytes.
+fn size_to_mb(size: i64, unit: TargetSizeUnit) -> Result<i64> {
+    if size <= 0 {
```
yhmo, line 735: src/v2/request/utility.rs line:717
Low ---- This rejects target_size(0), which the previous validation (target_size < 0) explicitly allowed and which meant "server default" on the wire (proto 0). Existing callers on the 2.6 branch that pass 0 will now fail at build time; they must omit the field instead. This matches pymilvus, but on a release branch it is a silent behavior change — consider documenting it in the PR description or keeping 0 as an alias for the server default.

src/v2/request/dql.rs
```diff
@@ -1597,10 +1597,10 @@ impl HybridSearchRequestBuilder {
...
         positive_i64("group_size", self.value.group_size)?;
         validate_search_extra_params(&self.value.extra_params)?;
-        if self.value.round_decimal < -1 {
+        if !(-1..=6).contains(&self.value.round_decimal) {
             return Err(Error::validation(
                 "round_decimal".into(),
-                "must be -1 or greater".into(),
+                "must be in -1..=6".into(),
             ));
         }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/utility.rs
```diff
@@ -624,12 +631,23 @@ impl CompactRequest {
...
 
+    /// Returns whether L0 compaction.
+    pub fn is_l0_compaction(&self) -> bool {
+        self.l0_compaction
+    }
+
+    /// Returns the unit used to interpret the target size.
+    pub fn target_size_unit(&self) -> TargetSizeUnit {
+        self.target_size_unit
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/utility.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/utility.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/utility.rs
```diff
@@ -668,22 +686,53 @@ impl CompactRequestBuilder {
...
+        .checked_mul(unit.bytes())
+        .ok_or_else(|| Error::validation("target_size".into(), "value is too large".into()))?;
+    let megabytes = bytes / BYTES_PER_MB;
+    if megabytes < 1 {
+        return Err(Error::validation(
+            "target_size".into(),
+            "must be at least 1MB".into(),
+        ));
+    }
+    Ok(megabytes)
```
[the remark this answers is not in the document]
yhmo: The truncation matches pymilvus `parse_target_size`, which also converts to MB with floor division (`int((value * unit_bytes) // BYTES_PER_MB)`), so this is parity-aligned rather than a divergence. GB/TB/PB are powers of 1024 and convert without truncation; only sub-MB KB/Byte values round, and any result below 1 MB is rejected instead of silently sending 0. Since `target_size` is a server-side target hint in MB granularity, no code change is needed.

PR #157, opened 2026-09-04
Add scripts/bump_version.sh to centralize future version bumps

yhmo:
Add scripts/bump_version.sh to centralize version bumps for releases.

The script updates, in one call:

- the root `Cargo.toml` package version;
- every `tutorial/*/Cargo.toml` exact `milvus-sdk-rust = "=X.Y.Z"` pin;
- every tutorial README version reference.

It validates the version format, is idempotent on re-run, and performs a final mismatch check across the root manifest, tutorial manifests, and tutorial READMEs so a release never ships with inconsistent pins.

Also update AGENTS.md to document the current active V2 feature areas and the decoding-only exception to the `new()` convention, and add yhmo to OWNERS reviewers and approvers.

### DIFF
--- a/scripts/bump_version.sh
+++ b/scripts/bump_version.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+# Licensed to the LF AI & Data foundation under one
+# or more contributor license agreements. See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership. The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License. You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Bump the crate version in Cargo.toml and sync the pinned milvus-sdk-rust
+# version across every standalone tutorial manifest and tutorial README.
+
+set -euo pipefail
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+MANIFEST="$ROOT_DIR/Cargo.toml"
+
+if [[ "$#" -ne 1 ]]; then
+  echo "Usage: $0 X.Y.Z" >&2
+  exit 1
+fi
+
+version="$1"
+if [[ ! "${version}" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+  echo "Invalid version: ${version}. Expected X.Y.Z." >&2
+  exit 1
+fi
+
+grep -q '^name = "milvus-sdk-rust"' "$MANIFEST" || {
+  echo "No milvus-sdk-rust manifest found at ${MANIFEST}." >&2
+  exit 1
+}
+
+# Portable in-place editing for both GNU and BSD sed: write to a temporary
+# file and move it over the original so `-i` flag differences never matter.
+# Exits 0 when the file actually changed, 1 when it was already up to date.
+sed_in_place() {
+  local file="$1"
+  shift
+  if sed "$@" "$file" > "${file}.tmp"; then
+    if ! cmp -s "$file" "${file}.tmp"; then
+      mv "${file}.tmp" "$file"
+      return 0
+    fi
+    rm -f "${file}.tmp"
+  fi
+  return 1
+}
+
+# README references look like `milvus-sdk-rust` version `X.Y.Z`.
+version_ref_pattern='`milvus-sdk-rust` version `[0-9]+\.[0-9]+\.[0-9]+`'
+target_ref="\`milvus-sdk-rust\` version \`${version}\`"
+
+updated=0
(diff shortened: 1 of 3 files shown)


scripts/bump_version.sh
```diff
@@ -0,0 +1,83 @@
...
+fi
+
+grep -q '^name = "milvus-sdk-rust"' "$MANIFEST" || {
+  echo "No milvus-sdk-rust manifest found at ${MANIFEST}." >&2
+  exit 1
+}
+
+updated=0
+
+if ! grep -q "^version = \"${version}\"" "$MANIFEST"; then
```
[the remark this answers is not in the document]
yhmo: Fixed. Version checks now use `grep -Fxq` (fixed-string, whole-line) instead of a regex with the version embedded, so dots in `3.0.0` are never treated as wildcards. Applied to the root manifest check, the tutorial manifest pin loop, and the final mismatch verification.

scripts/bump_version.sh
```diff
@@ -0,0 +1,83 @@
...
+
+for readme in "$ROOT_DIR"/tutorial/*/README.md; do
+  before="$(sha256sum "$readme" | cut -d' ' -f1)"
+  sed -i "s/\`milvus-sdk-rust\` version \`[0-9][0-9]*\.[0-9][0-9]*\.[0-9][0-9]*\`/\`milvus-sdk-rust\` version \`${version}\`/g" "$readme"
+  after="$(sha256sum "$readme" | cut -d' ' -f1)"
+  if [[ "$before" != "$after" ]]; then
+    echo "Updated ${readme#${ROOT_DIR}/} to milvus-sdk-rust ${version}"
+    updated=$((updated + 1))
+  fi
+done
```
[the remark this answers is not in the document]
yhmo: Fixed. Replaced GNU-specific `sed -i` and `sha256sum` with a portable `sed_in_place` helper that writes to a temp file and moves it over the original, so both GNU and BSD sed work on Linux and macOS. The README change detection now uses grep instead of hashing.

scripts/bump_version.sh
```diff
@@ -0,0 +1,93 @@
...
+
+if ! grep -Fxq "version = \"${version}\"" "$MANIFEST"; then
+  echo "Mismatch: root Cargo.toml is not set to version ${version}." >&2
+  exit 1
+fi
+for manifest in "$ROOT_DIR"/tutorial/*/Cargo.toml; do
```
yhmo, line 91: scripts/bump_version.sh line:82
Medium ---- The final mismatch check verifies the root Cargo.toml and every tutorial/*/Cargo.toml, but never the tutorial READMEs, even though the script's stated job (and this PR's title) is to sync the README version references too. If the README `sed` ever fails or the pattern stops matching (e.g. a README uses a slightly different phrasing), the script still exits 0 with "Version bumped to ... across N file(s)" while the README stays stale. Consider adding a README verification loop (e.g. assert that no `milvus-sdk-rust` version reference matching the old pattern remains) alongside the manifest checks.

scripts/bump_version.sh
```diff
@@ -0,0 +1,93 @@
...
+  if ! grep -Fxq "milvus-sdk-rust = \"=${version}\"" "$manifest"; then
+    sed_in_place "$manifest" "s/^milvus-sdk-rust = \"=.*\"$/milvus-sdk-rust = \"=${version}\"/"
+    echo "Updated ${manifest#${ROOT_DIR}/} to milvus-sdk-rust ${version}"
+    updated=$((updated + 1))
+  fi
+done
+
+for readme in "$ROOT_DIR"/tutorial/*/README.md; do
+  if grep -Eq "$version_ref_pattern" "$readme" && ! grep -Fq "$target_ref" "$readme"; then
```
yhmo: scripts/bump_version.sh line:71
Low ---- The README update guard skips the file entirely whenever the target reference is already present anywhere in it (`! grep -Fq "$target_ref"`). If a README already mentions the target version in one place but still has an older `milvus-sdk-rust` version reference elsewhere (e.g. a partially hand-edited tutorial README), the whole file is skipped and the stale reference survives undetected. The `sed` below is anchored to the backtick pattern and uses `/g`, so it is idempotent; consider running it whenever `grep -Eq "$version_ref_pattern"` matches and dropping the target-presence guard, then verifying no non-target reference remains.

yhmo:
AGENTS.md line:8
Low ---- The PR description says this PR "Bump[s] the root Cargo.toml package version from 2.6.0 to 3.0.0 and sync[s] the published crate version across every standalone tutorial manifest and tutorial README," but the diff contains no Cargo.toml or tutorial changes. Those updates already landed on `main` in "Prepare for v3.0.0 (#156)"; this PR only adds `scripts/bump_version.sh` and edits AGENTS.md. Consider updating the description so the actual scope is clear for release notes and reviewers.

scripts/bump_version.sh
yhmo: Fixed. Added a tutorial README verification loop to the final mismatch check: after replacement, any README that still matches the version-reference pattern without the target version fails the script with exit 1, alongside the existing root/tutorial manifest checks.

scripts/bump_version.sh
yhmo: Fixed. Dropped the target-presence guard so the README replacement runs whenever a version reference matches, normalizing partially hand-edited files. sed_in_place now uses cmp and returns whether the file actually changed, keeping the updated-file count and idempotent re-runs accurate; the new final README verification catches any stale reference that survives.

PR #158, opened 2026-09-04
Align run_analyzer JSON params and fill pymilvus parity gaps

yhmo:
Change RunAnalyzerRequest.analyzer_params from String to an optional serde_json::Value so callers pass a JSON object (matching the C++ and Java SDKs); into_proto serializes it back to the string wire field. This is an intentional incompatible API change.

Close pymilvus parity gaps reported against the main dev line:

- CreateSimpleCollectionRequest now accepts num_shards, num_partitions, properties, description, and id_type string aliases ("int"/"string"/"str").
- LoadCollectionRequest and LoadPartitionsRequest accept load_priority, forwarded as the load_priority load param.
- DmlResponse exposes cost() parsed from the server status report_value.
- describe_collection exposes consistency_level_name, CollectionSchema schema_version, FieldSchema field_id/is_dynamic/is_function_output, and Function id/input_field_ids/output_field_ids.

### DIFF
--- a/src/v2/request/collection.rs
+++ b/src/v2/request/collection.rs
@@ -162,16 +162,16 @@ impl From<CreateSimpleCollectionRequest> for CreateCollectionRequest {
         CreateCollectionRequest {
             database_name: value.database_name,
             collection_name: value.collection_name,
-            description: None,
+            description: value.description,
             schema: Some(schema),
-            num_partitions: 0,
-            num_shards: 1,
+            num_partitions: value.num_partitions,
+            num_shards: value.num_shards,
             consistency_level: value.consistency_level,
             index_params: vec![IndexParam::new()
                 .field_name(value.vector_field)
                 .index_type(IndexType::AutoIndex)
                 .metric_type(value.metric_type)],
-            properties: HashMap::new(),
+            properties: value.properties,
         }
     }
 }
@@ -314,12 +314,17 @@ pub struct CreateSimpleCollectionRequest {
     pub(crate) dimension: u32,
     pub(crate) primary_field: String,
     pub(crate) primary_field_type: DataType,
+    pub(crate) id_type: Option<String>,
     pub(crate) max_length: u32,
     pub(crate) vector_field: String,
     pub(crate) auto_id: bool,
     pub(crate) enable_dynamic_field: bool,
     pub(crate) consistency_level: ConsistencyLevel,
     pub(crate) metric_type: MetricType,
+    pub(crate) num_shards: i32,
+    pub(crate) num_partitions: i64,
+    pub(crate) properties: HashMap<String, String>,
+    pub(crate) description: Option<String>,
 }
 
 impl CreateSimpleCollectionRequest {
@@ -360,6 +365,11 @@ impl CreateSimpleCollectionRequest {
         self.primary_field_type
     }
 
+    /// Returns the raw `id_type` string alias, if one was set.
+    pub fn id_type(&self) -> Option<&str> {
+        self.id_type.as_deref()
+    }
+
     /// Returns the max length.
     pub fn max_length(&self) -> u32 {
         self.max_length
@@ -389,6 +399,26 @@ impl CreateSimpleCollectionRequest {
     pub fn metric_type(&self) -> MetricType {
         self.metric_type
     }
+
+    /// Returns the number of shards.
+    pub fn num_shards(&self) -> i32 {
+        self.num_shards
+    }
+
+    /// Returns the number of partitions.
+    pub fn num_partitions(&self) -> i64 {
+        self.num_partitions
+    }
+
+    /// Returns the collection properties.
+    pub fn properties(&self) -> &HashMap<String, String> {
+        &self.properties
+    }
+
+    /// Returns the collection description.
+    pub fn description(&self) -> Option<&str> {
+        self.description.as_deref()
+    }
 }
 
 impl CreateSimpleCollectionRequest {
@@ -399,12 +429,17 @@ impl CreateSimpleCollectionRequest {
             dimension: 0,
             primary_field: "id".into(),
             primary_field_type: DataType::Int64,
+            id_type: None,
             max_length: 65_535,
             vector_field: "vector".into(),
             auto_id: false,
             enable_dynamic_field: true,
             consistency_level: ConsistencyLevel::Bounded,
             metric_type: MetricType::Cosine,
+            num_shards: 1,
+            num_partitions: 0,
+            properties: HashMap::new(),
+            description: None,
         }
     }
 }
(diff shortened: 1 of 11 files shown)


src/v2/response/dml.rs
```diff
@@ -114,8 +132,15 @@ impl DmlResponse {
             delete_count: value.delete_cnt,
             upsert_count: value.upsert_cnt,
             timestamp: value.timestamp,
+            cost: parse_extra(&extra_info, "report_value", -1_i64),
         })
     }
+
+    /// Returns the cost in milliseconds reported by the server, or `-1` when
```
yhmo: src/v2/response/dml.rs line:139
Medium ---- The doc comment says the cost is "in milliseconds", but the value parsed here is the server's `report_value` status field. pymilvus documents the same field with "The unit of this cost is vcu, similar to token" (see pymilvus/orm/mutation.py), and the sibling `SearchResponse::cost()` doc in dql.rs deliberately omits a unit. If report_value is a VCU-style cost rather than an elapsed time, "milliseconds" would mislead callers — please verify the unit or drop the unit claim.

src/v2/response/dml.rs
```diff
@@ -20,6 +20,16 @@ use crate::proto::milvus;
 use crate::v2::error::Result;
 pub use crate::v2::types::Ids;
 
+fn parse_extra<T>(values: &std::collections::HashMap<String, String>, key: &str, default: T) -> T
```
yhmo: src/v2/response/dml.rs line:23
Low ---- `parse_extra` here is byte-for-byte identical to the helper already defined in src/v2/response/dql.rs:1725. Consider extracting one shared extra-info parser instead of duplicating it, so future changes to how `report_value`/`scanned_*`/`cache_hit_ratio` are decoded do not drift across the two response modules.

src/v2/request/collection.rs
```diff
@@ -485,33 +520,93 @@ impl CreateSimpleCollectionRequestBuilder {
         self
     }
 
+    /// Sets the primary id type and returns the updated value.
+    ///
+    /// Accepts `"int"` (maps to [`DataType::Int64`]) and `"string"`/`"str"`
+    /// (maps to [`DataType::VarChar`]), matching pymilvus `create_collection`.
+    /// The alias is resolved during [`CreateSimpleCollectionRequestBuilder::build`].
+    pub fn id_type(mut self, value: impl Into<String>) -> Self {
```
yhmo, line 528: src/v2/request/collection.rs line:528
Medium ---- The new CreateSimpleCollectionRequest surface (id_type alias, num_shards, num_partitions, properties, description) is not exercised by any unit test: create_simple_collection_request_default_values/populated_values do not set or assert these fields, and build() has no test for the id_type alias resolution ("int"/"string"/"str"), the num_shards<=0 / num_partitions<0 validation, or the From<CreateSimpleCollectionRequest> forwarding. load_priority forwarding and DmlResponse cost parsing from report_value are likewise untested (dql.rs has a report_value test; dml.rs does not). A few builder/validation tests would lock in the pymilvus-parity behavior this PR is adding.

src/v2/response/dml.rs
yhmo: Fixed. Dropped the unit claim from the cost doc and made it consistent with SearchResponse::cost() (which deliberately omits a unit). pymilvus documents report_value as a VCU-style cost, not elapsed milliseconds, so the previous wording was misleading.

src/v2/response/dml.rs
yhmo: Fixed. Extracted the duplicated parse_extra into a shared pub(crate) helper in src/v2/utils.rs and updated both src/v2/response/dml.rs and src/v2/response/dql.rs to use it, so report_value/scanned_*/cache_hit_ratio decoding no longer drifts across modules.

src/v2/request/collection.rs
yhmo: Fixed. Added unit tests covering the new CreateSimpleCollectionRequest surface: id_type alias resolution (int/string/str + invalid), num_shards<=0 and num_partitions<0 validation, and From<CreateSimpleCollectionRequest> field forwarding; load_priority forwarding (getter + into_proto load_params) for both load_collection and load_partitions; and DmlResponse cost parsing from report_value (plus the -1 default when absent).

src/v2/request/collection.rs
```diff
@@ -360,6 +365,11 @@ impl CreateSimpleCollectionRequest {
         self.primary_field_type
     }
 
+    /// Returns the raw `id_type` string alias, if one was set.
+    pub fn id_type_alias(&self) -> Option<&str> {
```
yhmo: src/v2/request/collection.rs line:369
Low ---- The `id_type_alias()` accessor can never observe a value on a built request: `build()` consumes the alias via `id_type_alias.take()` to resolve `primary_field_type`, so the accessor always returns `None` after a successful build. That makes the doc comment ("if one was set") and the setter/accessor naming asymmetry (`id_type()` setter vs `id_type_alias()` accessor, which also forced the special-case `"id_type_alias": ("id_type",)` entry in `scripts/check_v2_api.py`) misleading for callers. Consider naming the field/accessor `id_type` (returning `Option<&str>`) so setter and accessor match, or dropping the accessor in favor of the already-resolved `primary_field_type()`.

src/v2/request/utility.rs
```diff
@@ -1110,8 +1114,8 @@ pub struct RunAnalyzerRequestBuilder {
 
 impl RunAnalyzerRequestBuilder {
     /// Sets the analyzer params and returns the updated value.
-    pub fn analyzer_params(mut self, value: impl Into<String>) -> Self {
-        self.value.analyzer_params = value.into();
+    pub fn analyzer_params(mut self, value: serde_json::Value) -> Self {
```
yhmo, line 1117: src/v2/request/utility.rs line:1117
Low ---- `analyzer_params` accepts any `serde_json::Value`, but the parameter is documented and exemplified (examples/v2/run_analyzer.rs) as a JSON object. A caller can pass `json!(42)` or `json!("...")`, which `into_proto` serializes to a non-object string that the server rejects only at RPC time. Consider validating `value.is_object()` in `build()` so invalid params fail fast with a typed error, consistent with the other builder validations (and with the SDK's stated "report invalid or conflicting parameters before an RPC is sent" contract).

src/v2/types/collection.rs
```diff
@@ -1153,6 +1192,11 @@ impl CollectionSchema {
         self.enable_dynamic_field
     }
 
+    /// Returns the schema version.
+    pub fn get_schema_version(&self) -> i32 {
```
yhmo, line 1196: src/v2/types/collection.rs line:1196
Low ---- The PR description states describe_collection now exposes "CollectionSchema enable_namespace/schema_version", but only `schema_version` is implemented here: `CollectionSchema` has no `enable_namespace` member, and the existing `namespace_is_not_exposed_or_preserved_by_sdk_schema` test asserts the round-trip intentionally drops it. Either correct the description, or treat `enable_namespace` as a remaining parity gap against the referenced pymilvus `describe_collection` output.

src/v2/request/collection.rs
yhmo: Fixed. Renamed the field and accessor from id_type_alias to id_type, so the setter (id_type) and accessor (id_type) now match and the special-case "id_type_alias": ("id_type",) entry in scripts/check_v2_api.py is removed. The accessor reflects the raw alias before build() resolves it into primary_field_type.

src/v2/request/utility.rs
yhmo: Fixed. build() now validates that analyzer_params, when present, is a JSON object (value.is_object()); non-object values fail fast with a typed validation error before the RPC, consistent with the SDK's fail-fast contract. Added a unit test covering string and number rejection.

src/v2/types/collection.rs
yhmo: Fixed. The PR description was updated to remove enable_namespace (experimental, intentionally not exposed in the Rust SDK per maintainer direction); describe_collection now exposes CollectionSchema schema_version only, matching the implementation.

PR #160, opened 2026-09-08
Support search/query ORDER BY fields, optimize task status

yhmo:


src/v2/response/collection.rs
```diff
@@ -322,13 +324,29 @@ impl GetCollectionStatsResponse {
         }
     }
 
+    /// Returns the name of the collection the statistics describe.
```
yhmo, line 327: src/v2/response/collection.rs line:327
Low ---- This PR bundles three unrelated changes: ORDER BY support for search/query, `OptimizeTask::task_status()`, and these `name()`/`row_count()` additions to the collection and partition statistics responses. The statistics-response additions are not mentioned in the PR title or body. Consider splitting them into a separate PR, or documenting the rationale here, so each change stays independently reviewable and revertible.

src/v2/request/dql.rs
```diff
@@ -2110,6 +2163,32 @@ fn set_search_param(params: &mut Vec<common::KeyValuePair>, key: &str, value: im
...
+            return Err(Error::validation(
+                "order_by_fields".into(),
+                "field name cannot be empty".into(),
+            ));
+        }
+    }
+    Ok(())
+}
+
+fn format_order_by_fields(order_by_fields: &[OrderByField]) -> String {
```
yhmo, line 2179: src/v2/request/dql.rs line:2178
Low ---- ORDER BY is currently covered only by wire-encoding unit tests (`into_proto` assertions) and mock-based v2_ut tests; there is no v2_st system test or example that exercises `order_by_fields` against a real server. Direction strings (`asc`/`desc`), sortable-field validation, and the iterator/group-by restrictions are all enforced server-side (see parseOrderByFields), so a format or semantics mismatch would only surface in a system test. Non-blocking follow-up: add one once a server with ORDER BY support is available.

src/v2/client/utility.rs
```diff
@@ -110,6 +110,15 @@ impl OptimizeTask {
...
     }
+
+    /// Returns the final task status, or success while the task is still running.
+    pub fn task_status(&self) -> Result<()> {
+        match self.state.result.read().as_ref() {
+            Some(Ok(_)) => Ok(()),
+            Some(Err(error)) => Err(error.clone()),
+            None => Ok(()),
+        }
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
```diff
@@ -871,6 +896,13 @@ impl SearchRequest {
                 set_search_param(&mut search_params, "metric_type", metric.as_str());
             }
         }
+        if !self.order_by_fields.is_empty() {
+            set_search_param(
+                &mut search_params,
+                "order_by_fields",
+                format_order_by_fields(&self.order_by_fields),
+            );
+        }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/response/collection.rs
yhmo: as design

src/v2/response/collection.rs
```diff
@@ -322,13 +324,29 @@ impl GetCollectionStatsResponse {
...
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    /// Returns the server-reported row count, when present.
+    pub fn row_count(&self) -> Option<i64> {
+        self.statistics
+            .get("row_count")
+            .and_then(|value| value.parse::<i64>().ok())
```
yhmo, line 340: src/v2/response/collection.rs line:336
Low ---- `row_count()` silently swallows a malformed `row_count` stat: if the server returns a non-numeric value, callers get `None` exactly as if the stat were absent, so a server-side data anomaly becomes indistinguishable from a missing value. The same lenient parse is duplicated in `src/v2/response/partition.rs:69`. The repo's response convention (AGENTS.md: "never fabricate defaults or silently drop malformed fields") points toward surfacing this rather than hiding it; the raw value remains reachable via `statistics()`, so consider returning a typed parse error or documenting the lenient fallback on these new public accessors.

yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
```diff
@@ -177,6 +184,12 @@ impl QueryRequest {
         if !self.timezone.is_empty() {
             params.insert("timezone".into(), self.timezone);
         }
+        if !self.order_by_fields.is_empty() {
+            params.insert(
+                "order_by_fields".into(),
```
yhmo, line 189: src/v2/request/dql.rs line:189
Low ---- If a caller sets both the structured `order_by_fields` member and a raw `extra_params` entry whose key is `order_by_fields`, this `insert` silently discards the raw value. On SearchRequest the same combination is rejected at build time because the key is reserved in `validate_search_extra_params`. The raw passthrough on query is a working legacy path (it reaches `query_params` unchanged when the structured member is empty), so I am not suggesting removing it — but please confirm the silent-override precedence is intended, or reject the conflicting combination so a raw value is never discarded without notice.

src/v2/request/dql.rs
```diff
@@ -1896,12 +1941,19 @@ fn validate_query_iterator_query(value: &QueryRequest) -> Result<()> {
...
+            "query iterator does not support ORDER BY".into(),
+        ));
+    }
     validate_query_request_limit(value, true)
 }
 
 fn validate_query_request_limit(value: &QueryRequest, allow_zero_limit: bool) -> Result<()> {
     required("collection_name", &value.collection_name)?;
     non_empty_strings("partition_names", &value.partition_names)?;
+    validate_order_by_fields(&value.order_by_fields)?;
```
yhmo, line 1956: src/v2/request/dql.rs line:1956
Low ---- The server requires an explicit limit whenever ORDER BY is used (the proxy rejects with "ORDER BY requires explicit limit" when no limit param is present, to bound segment-level sort memory). QueryRequest currently allows building with `order_by_fields` set and `limit` omitted; the failure only surfaces as a server-side error at query time. Consider validating in `validate_query_request_limit` that `limit` is present when `order_by_fields` is non-empty, matching the server contract and the iterator-side guard already added in this PR.

src/v2/request/dql.rs
yhmo: The silent-override precedence is intentional to align with the C++ and Java SDKs: C++ has no raw query-param passthrough for order_by_fields (the typed member is the only route), and Java forwards both the raw map value and the typed order_by_fields as duplicate key-value pairs with the typed value appended last. Rust's query path keeps a single value (typed wins) and its search path already rejects the conflicting combination at build time. No change is planned.

src/v2/request/dql.rs
yhmo: Keeping the current behavior to align with the C++ and Java SDKs: neither C++ QueryRequest nor Java QueryReq validates the ORDER BY + explicit-limit combination client-side. Both rely on the server-side 'ORDER BY requires explicit limit' rejection (internal/proxy/task_query.go), exactly as Rust does today. No change is planned.

.github/mergify.yml
```diff
@@ -39,9 +39,12 @@ pull_request_rules:
       - 'base~=^(main|[0-9]+[.][0-9]+)$'
       - label=ci-passed
       - or:
-        - "-check-success=Build and test on Ubuntu-22.04"
-        - "-check-success=Build and test on MacOS 15"
-        - "-check-success=Build and test on Windows"
+          - "check-pending=Build and test on Ubuntu-22.04"
```
yhmo, line 42: .github/mergify.yml line:42
Low ---- The previous rule used `-check-success=<job>` (match when a check is anything but success), so an errored/timed-out/cancelled/neutral check removed `ci-passed`. The new conditions match only `check-pending` or `check-failure`, so a check that ends in `error`, `timed_out`, `cancelled`, or `neutral` no longer removes the label -- the merge gate would consider CI green even though one of the three platform checks did not pass. If the intent is "any not-green check removes ci-passed", keep a `-check-success` arm (or also match the error/neutral states) alongside the pending/failure conditions.

yhmo: The concern is already covered by Mergify's documented check-failure semantics: per docs.mergify.com/conditions, checks that report being cancelled, timed out, or action_required are also considered failures by check-failure, so those states do remove ci-passed. GitHub Actions jobs end in success, failure, or cancelled, and the three platform jobs here run unconditionally (no path filter), so neutral/skipped/stale never occur for them. This mirrors the production C++ SDK config, so the change is kept as-is.

PR #164, opened 2026-09-10
Remove cargo package staging and pycache artifacts from release publishing

yhmo:


Cargo.toml
```diff
@@ -22,6 +22,7 @@ include = [
     "/CHANGELOG.md",
     "/LICENSE",
 ]
+exclude = ["**/__pycache__/**"]
```
yhmo: Cargo.toml line:25
Medium ---- The new `exclude = ["**/__pycache__/**"]` has no effect: this manifest already sets `include = [...]`, and Cargo treats `include`/`exclude` as mutually exclusive with `include` overriding `exclude` (Cargo Book "The `exclude` and `include` fields"; confirmed in cargo 1.86 `src/cargo/sources/path.rs` `_list_files`, where `exclude` rules are consulted only when `include` is empty). Because `include` is set, `.gitignore` is also not applied to packaging, so a `__pycache__/` directory under `/tests/**` would still be shipped in the published crate if it exists at publish time. If the goal is to keep pycache out of the crate, add a negation to the `include` list (e.g. `"!/**/__pycache__/**"` or `"!**/*.py[cod]"`) instead; as written, this line is dead config that can mislead future maintainers.

.github/workflows/publish-release.yml
```diff
@@ -171,6 +171,13 @@ jobs:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
         run: cargo publish
 
+      # `cargo publish --dry-run` stages the package under target/package. Swatinem/rust-cache's
```
yhmo, line 174: .github/workflows/publish-release.yml line:174
Low ---- Question on the mechanism in this comment: I read Swatinem/rust-cache v2.7.7 `src/cleanup.ts` and its `cleanTargetDir` only recurses into directories detected as nested targets (those containing `CACHEDIR.TAG` or `.rustc_info.json`). `target/package` has neither, so it is treated as a plain profile dir and the walk aborts (and is swallowed) at the first missing `build`/`.fingerprint`/`deps` entry before ever reaching `target/package/*/tests/{target,trybuild}`. Also, `main.yml` `ut-linux` runs `cargo package` (which stages the same `target/package`) with the same `rust-cache@v2` key and has no such workaround. Could you confirm which rust-cache version/log produced the ENOENT annotations? `rm -rf target/package` is harmless regardless, but as written the comment may mislead a future maintainer debugging this workflow.

.github/workflows/publish-release.yml
```diff
@@ -171,6 +171,13 @@ jobs:
...
+      # post-job save walks target/ recursively and treats any `tests` directory as a trybuild
+      # profile, so it tries to opendir target/package/*/tests/{target,trybuild} and emits
+      # ENOENT annotations. Remove the staging dir before the post-job cache save runs.
+      - name: Remove staged package before cache save
+        run: rm -rf target/package
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

Cargo.toml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/publish-release.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/publish-release.yml
```diff
@@ -171,6 +171,15 @@ jobs:
...
         run: cargo publish
 
+      # `cargo publish --dry-run` stages the package under target/package. Swatinem/rust-cache's
+      # post-job save (v2.9.x, after the 2026-08-06 "invert target/profile check" change) walks
+      # target/ recursively and treats any `tests` directory as a trybuild profile, so it tries to
+      # opendir target/package/*/tests/{target,trybuild} and emits ENOENT annotations. Remove the
+      # staging dir before the post-job cache save runs, even when publish itself failed.
+      - name: Remove staged package before cache save
+        if: always()
+        run: rm -rf target/package
```
yhmo, line 181: .github/workflows/publish-release.yml line:181
Medium ---- The `.github/workflows/main.yml` `ut-linux` job combines the same two ingredients this step guards against: `Swatinem/rust-cache@v2` (main.yml:59) and `cargo package` (main.yml:84). `cargo package` verifies by default and unpacks the crate to `target/package/milvus-sdk-rust-<version>/` (including `tests/`), which it leaves in place, so the job's post-step cache save walks the same staged `tests` directory on every push/PR. Because both workflows resolve the same floating `@v2` action, rust-cache's save-side cleanup runs the same `tests`-as-trybuild traversal there. Was the `ut-linux` exposure considered? If the failure is real in the publish job it is real in `ut-linux` too; either both jobs should remove `target/package` (or a shared helper should do it), or the annotation/cleanup there is intentionally tolerated and worth documenting.

yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #165, opened 2026-09-10
Support client-side bloom_match and roaring_match membership-filter blobs (#131)

yhmo:
issue #131
Add MBF1 (Split-Block Bloom Filter) and MRB1 (portable Roaring64) blob builders under milvus::v2 and wire them into V2 query, search, hybrid search (sub-search), and delete requests through a bytes filter-template channel that sends the blob as TemplateValue.bytes_val. Both builders reproduce the shared cross-SDK golden vectors byte-for-byte.


src/v2/request/dml.rs
```diff
@@ -580,6 +594,31 @@ pub(crate) fn json_template(value: Value) -> Result<crate::proto::schema::Templa
...
+    json_templates: HashMap<String, Value>,
+    bytes_templates: HashMap<String, Vec<u8>>,
+) -> Result<HashMap<String, crate::proto::schema::TemplateValue>> {
+    use crate::proto::schema::{template_value, TemplateValue};
+    let mut values = json_templates
+        .into_iter()
+        .map(|(key, value)| Ok((key, json_template(value)?)))
+        .collect::<Result<HashMap<String, TemplateValue>>>()?;
+    for (key, blob) in bytes_templates {
+        values.insert(
```
yhmo, line 635: src/v2/request/dml.rs line:612
Low ---- `template_values` silently overwrites a JSON template with a bytes template when the same key is present in both maps: the JSON map is built first (lines 607-610) and the loop here unconditionally inserts the bytes blob, so a caller who sets a key through both `filter_templates(...)` and the new `filter_template_bytes(...)` gets the bytes blob with no error. A stale bytes entry carried over through `into_builder()`/re-build can also silently mask an intended JSON template. Since the SDK rejects ambiguous inputs elsewhere at build time, consider returning a validation error when a key appears in both maps, or at least document the precedence in both builder methods.

src/v2/bloom_filter.rs
```diff
@@ -0,0 +1,343 @@
...
+        })
+    }
+
+    /// Returns the number of 32-byte blocks in the filter body.
+    pub fn num_blocks(&self) -> u32 {
+        self.num_blocks
+    }
+
+    /// Inserts an int64 value (8-byte little-endian encoding).
+    pub fn add_int64(&mut self, v: i64) {
```
yhmo: src/v2/bloom_filter.rs line:182
Low ---- The two new public blob builders are inconsistent with each other and with the codebase fluent style. `BloomFilterBuilder::add_int64` / `add_string` return `()` (non-chainable) and the terminal call is `marshal()` returning `Vec<u8>`, while `RoaringBitmapBuilder::add_int64` / `add_int64s` return `&mut Self` and the terminal call is `build()` returning `Result<Vec<u8>>`. V2 convention favors consuming, chainable setters, and the roaring side already follows it within this same PR. Aligning the bloom builder (chainable `add_*` returning `&mut Self` or `Self`, and one consistent terminal method name) would make the two modules behave predictably for callers.

src/v2/request/dml.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/bloom_filter.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/bloom_filter.rs
```diff
@@ -0,0 +1,345 @@
...
+fn block_index(hash: u64, num_blocks: u32) -> usize {
+    (((hash >> 32) * num_blocks as u64) >> 32) as usize
+}
+
+/// Incrementally constructs a Split-Block Bloom Filter and serializes it into an MBF1 envelope.
+///
+/// The filter is sized from `n` and `fpr` alone — never from the values — so members can be added
+/// in any order, duplicated, or streamed in one at a time. Not safe for concurrent use.
+#[derive(Debug, Clone)]
+pub struct BloomFilterBuilder {
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
```diff
@@ -284,6 +288,16 @@ impl QueryRequestBuilder {
         self
     }
 
+    /// Adds a bytes filter template — a client-built membership-filter blob (e.g. a
+    /// [`crate::v2::bloom_filter::BloomFilterBuilder`] or
+    /// [`crate::v2::roaring_bitmap::RoaringBitmapBuilder`] blob) for a `bloom_match` /
+    /// `membership_match` expression — and returns the updated value. The blob is sent as native
+    /// protobuf `TemplateValue.bytes_val`, never base64-inflated through a string field.
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/roaring_bitmap.rs
```diff
@@ -0,0 +1,583 @@
...
+struct Layout {
+    groups: Vec<GroupPlan>,
+    containers: Vec<ContainerPlan>,
+    /// An empty member set still writes the 8-byte high-container count, so the body is never 0.
+    body_length: u64,
+}
+
+/// Statistics about a built (or planned) roaring bitmap.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct RoaringBitmapStats {
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/roaring_bitmap.rs
```diff
@@ -0,0 +1,589 @@
...
+    /// An empty member set still writes the 8-byte high-container count, so the body is never 0.
+    body_length: u64,
+}
+
+///////////////////////////////////////////////////////////////////////////////
+// RoaringBitmapStats
+///////////////////////////////////////////////////////////////////////////////
+/// Statistics about a built (or planned) roaring bitmap.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct RoaringBitmapStats {
```
yhmo, line 122: src/v2/roaring_bitmap.rs line:122
Low ---- `RoaringBitmapStats` is the only public struct in `src/v2` whose fields are public; every other V2 public struct keeps fields private and exposes read-only accessors (see `src/v2/types/common.rs` `RetryConfig` and `src/v2/types/aggregation.rs` `MetricSpec`). An 8-field `#[derive(Clone, Copy)]` with `pub` fields diverges from the repository's V2 code style and makes any future field-access change a semver-visible break. Consider private fields with `field()`/`get_*()` accessors, or a doc note explaining the intentional deviation if this record-like shape is deliberate.

src/v2/roaring_bitmap.rs
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    /// Returns the number of distinct members.
+    pub fn cardinality(&mut self) -> u64 {
+        self.normalize();
+        self.keys.len() as u64
+    }
+
+    /// Returns statistics about the bitmap that would be built.
+    pub fn stats(&mut self) -> RoaringBitmapStats {
```
yhmo: src/v2/roaring_bitmap.rs line:190
Low ---- `stats()` runs `plan_layout()` directly with no size guard, so for a very large sparse member set it allocates one `ContainerPlan` per distinct low container before any check; `build()` and `validate()` call `check_bucket_limits()` first and reject the same input without allocating. For a set just past the decoded-size limit, `stats()` can transiently allocate hundreds of MB while the guarded `build()` path errors cleanly. Since `stats()` is documented as the pre-build size check, consider running the same bucket/limit pre-checks here (or a counted pass) or documenting the allocation cost for oversized inputs.

src/v2/roaring_bitmap.rs
```diff
@@ -0,0 +1,589 @@
...
+    out[*pos..*pos + 4].copy_from_slice(&value.to_le_bytes());
+    *pos += 4;
+}
+
+fn write_u64_le(out: &mut [u8], pos: &mut usize, value: u64) {
+    out[*pos..*pos + 8].copy_from_slice(&value.to_le_bytes());
+    *pos += 8;
+}
+
+#[cfg(test)]
```
yhmo, line 548: src/v2/roaring_bitmap.rs line:506
Medium ---- The memory guardrails that this feature depends on — `check_high_container_limit`, `check_bucket_limits`, `check_limits`, and `validate()` — have no test coverage; only the golden-vector round-trip, dedup, and sign-extension paths are tested. Since the whole point of roaring blobs is shipping very large membership sets, the reject-before-allocate behavior is exactly what should be pinned (mirror the Go suite's `TestBuildRejectsSparseHighContainersBeforeBitmapConstruction`: `MAX_HIGH_CONTAINER_COUNT + 1` members spread over distinct high groups, plus a set that exceeds `MAX_ESTIMATED_DECODED_BYTES`). These are cheap to add and would guard against a future regression that turns an oversized build into a large transient allocation instead of a clean error.

src/v2/request/dql.rs
```diff
@@ -228,10 +235,7 @@ impl QueryRequest {
                 .map(|level| level.into_proto() as i32)
                 .unwrap_or_default(),
             use_default_consistency: self.consistency_level.is_none(),
-            expr_template_values: filter_templates
-                .into_iter()
-                .map(|(key, value)| Ok((key, json_template(value)?)))
-                .collect::<Result<_>>()?,
+            expr_template_values: template_values(filter_templates, self.filter_template_bytes)?,
```
yhmo: src/v2/request/dql.rs line:238
Medium ---- In the ids branch (lines 200-214) `filter_templates` (the JSON templates) are deliberately discarded and replaced by the pk-in template, but `filter_template_bytes` is still passed unconditionally to `template_values`, so a request built with both `.ids(...)` and `.filter_template_bytes(...)` transmits the blob in `expr_template_values` even though the rewritten expression `{pk} in {__milvus_v2_query_ids}` never references it. The server may reject an unused template value, and the behavior diverges from the JSON path and from `DeleteRequest` (which drops both in its ids branch). Consider dropping `filter_template_bytes` in the ids branch too, or rejecting the combination at build time.

src/v2/request/dql.rs
```diff
@@ -284,6 +288,16 @@ impl QueryRequestBuilder {
         self
     }
 
+    /// Adds a bytes filter template — a client-built membership-filter blob (e.g. a
+    /// [`crate::v2::bloom_filter::BloomFilterBuilder`] or
+    /// [`crate::v2::roaring_bitmap::RoaringBitmapBuilder`] blob) for a `bloom_match` /
+    /// `membership_match` expression — and returns the updated value. The blob is sent as native
+    /// protobuf `TemplateValue.bytes_val`, never base64-inflated through a string field.
+    pub fn filter_template_bytes(mut self, key: impl Into<String>, blob: Vec<u8>) -> Self {
```
yhmo: src/v2/request/dql.rs line:296
Low ---- The builder method `filter_template_bytes(key, blob)` adds a single key/blob entry but is named like a whole-map setter, while the existing `filter_templates(mut self, value: HashMap)` replaces the entire map, and the accessor `filter_template_bytes()` returns the whole `&HashMap`. A caller seeing `filter_templates(...)` setter-style naming may reasonably expect `filter_template_bytes(HashMap<String, Vec<u8>>)` to set the whole bytes map in one call (mirroring `filter_templates`), or expect the method to return a value they can chain into another template — instead it only inserts one entry. Consider `add_filter_template_bytes(key, blob)` (matching the V1 `add_template_value` pattern) or a setter overload for symmetry, and apply the same naming across `SearchRequest`, `SubSearchRequest`, and `DeleteRequest`.

src/v2/roaring_bitmap.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/roaring_bitmap.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/roaring_bitmap.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/roaring_bitmap.rs
```diff
@@ -0,0 +1,654 @@
...
+
+impl RoaringBitmapBuilder {
+    /// Returns an empty builder.
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Inserts an int64 member. Signed values preserve their two's-complement bits when mapped
+    /// into the uint64 key space: `INT8(-1)` becomes `0xffffffffffffffff`, not `0xff`.
+    pub fn add_int64(&mut self, value: i64) -> &mut Self {
```
yhmo, line 199: src/v2/roaring_bitmap.rs line:196
Low ---- The builder accepts streamed `add_int64` calls but buffers every member as a `u64` in `keys` and only sorts/dedups in `build()`, so a large membership set still accumulates the full list in memory (10M members ≈ 80 MB of keys) before any blob is produced. This matches the Go reference, which also materializes its input slice, so it is not a correctness bug; the concern is the contract versus the doc: `RoaringBitmapBuilder` says "Incrementally builds an MRB1 roaring bitmap blob" and exposes a streaming `add_int64` entry point, which implies a set that does not fit in memory as a list can be fed in without materializing it (the exact scenario #131 was written to enable). Either document that members are buffered until `build()`, or build into a bitset incrementally so very large member sets do not double-hold the list in RAM.

src/v2/bloom_filter.rs
```diff
@@ -0,0 +1,348 @@
...
+            let word = u32::from_le_bytes(block[i * 4..i * 4 + 4].try_into().unwrap());
+            block[i * 4..i * 4 + 4].copy_from_slice(&(word | mask).to_le_bytes());
+        }
+    }
+
+    /// Stamps the MBF1 header onto the filter and returns the envelope.
+    ///
+    /// Calling `build` does not consume the builder, so more values may be inserted afterwards;
+    /// each call re-stamps the header and returns a fresh copy of the envelope.
+    pub fn build(&self) -> Vec<u8> {
```
yhmo: src/v2/bloom_filter.rs line:219
Low ---- `build()` clones the entire filter buffer on every call, and that buffer can be up to `MAX_FILTER_BYTES` (128 MiB). For a filter near the cap this briefly doubles peak memory. The Go reference avoids the copy by returning its buffer directly (documented read-only), whereas taking `&self` here forces the clone so `build()` can be called repeatedly. Non-blocking: the fresh-copy behavior is documented and safe, but a consuming `build(self) -> Vec<u8>` (or a variant that writes into a caller-owned buffer) would remove the extra allocation for the common build-once path.

src/v2/roaring_bitmap.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/bloom_filter.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/roaring_bitmap.rs
```diff
@@ -0,0 +1,657 @@
...
+    /// Returns the number of distinct members.
+    pub fn cardinality(&mut self) -> u64 {
+        self.normalize();
+        self.keys.len() as u64
+    }
+
+    /// Returns statistics about the bitmap that would be built. Runs the same bucket pre-checks as
+    /// [`RoaringBitmapBuilder::build`] first, so an oversized member set is rejected without
+    /// allocating a container plan per value.
+    pub fn stats(&mut self) -> Result<RoaringBitmapStats> {
```
yhmo, line 229: src/v2/roaring_bitmap.rs line:229
Medium ---- `stats()` is a weaker guardrail than `build()`/`validate()`: it runs `check_bucket_limits` (overhead derived from high/low container counts) but skips `check_limits`, which additionally rejects `estimated_decoded_size > MAX_ESTIMATED_DECODED_BYTES` and `body_length > MAX_BODY_BYTES`. A member set that is bucket-legal but has a large encoded body (e.g. many full bitmap containers) therefore makes `stats()` return Ok with a body_length while the very same set fails in `build()` — so a caller that uses `stats()` to pre-flight a blob gets a false green light. Please call `check_limits` from `stats()` as well (it already has the Layout to compute stats_of), or document the divergence.

src/v2/bloom_filter.rs
```diff
@@ -0,0 +1,348 @@
...
+            "fpr".into(),
+            format!("bloom filter fpr {fpr} out of range [{MIN_FPR}, {MAX_FPR}]"),
+        ));
+    }
+    Ok(BLOOM_HEADER_SIZE + optimal_num_of_bytes(n, fpr))
+}
+
+/// Builds an MBF1 bloom blob from an int64 membership set at false-positive rate `fpr`. `fpr`
+/// must lie in `[MIN_FPR, MAX_FPR]`; pass [`DEFAULT_FPR`] when you have no specific target.
+pub fn bloom_filter_blob_int64(members: &[i64], fpr: f64) -> Result<Vec<u8>> {
```
yhmo, line 253: src/v2/bloom_filter.rs line:248
Medium ---- The convenience helpers `bloom_filter_blob_int64`/`bloom_filter_blob_string` (and `roaring_bitmap_blob`) size the filter from `members.len()` and allocate the full body immediately (up to `MAX_FILTER_BYTES` = 128 MiB) with no pre-flight size guard, even though `estimate_size()` exists precisely so callers can reject an over-large filter before building. The MBF1 docstring motivates these blobs as a way to pass the proxy gRPC receive limit that a raw `in [...]` list would exceed, so a large membership set builds a multi-hundred-MB allocation and a blob that can itself exceed the transport limit. Consider routing the helpers through `estimate_size`/the bucket limits first (as `RoaringBitmapBuilder::stats` does) so oversized sets fail fast, or document the expected ceiling.

src/v2/roaring_bitmap.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/bloom_filter.rs
yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dml.rs
```diff
@@ -580,6 +594,40 @@ pub(crate) fn json_template(value: Value) -> Result<crate::proto::schema::Templa
...
+/// wire `expr_template_values` map. Bytes templates carry client-built membership-filter blobs
+/// (for example a [`crate::v2::bloom_filter::BloomFilterBuilder`] or
+/// [`crate::v2::roaring_bitmap::RoaringBitmapBuilder`] blob) as native
+/// `TemplateValue.bytes_val`, never base64-inflated through a string field.
+pub(crate) fn template_values(
+    json_templates: HashMap<String, Value>,
+    bytes_templates: HashMap<String, Vec<u8>>,
+) -> Result<HashMap<String, crate::proto::schema::TemplateValue>> {
+    use crate::proto::schema::{template_value, TemplateValue};
+    if let Some(key) = json_templates
```
yhmo: src/v2/request/dml.rs line:607
Low ---- The duplicate-template-key conflict is only validated here, inside `template_values`, which runs at request->proto conversion time (QueryRequest/SearchRequest/SubSearchRequest/DeleteRequest `into_proto`). Per the V2 request pattern, `build()` should validate cross-field constraints (QueryRequest already rejects ids+filter together in `validate_query_request`), so a caller who sets the same key through both `filter_templates(...)` and `add_filter_template_bytes(...)` gets a clean `build()` and only discovers the Validation error when the client call converts the request. Consider running the same key-conflict check in each builder's `build()` (or in the shared validation helpers) so conflicting inputs fail fast at construction like the other cross-field rules.

yhmo: The issue is confirmed fixed. Resolving this conversation.

src/v2/request/dql.rs
```diff
@@ -279,11 +291,31 @@ impl QueryRequestBuilder {
...
 
     /// Sets the filter templates and returns the updated value.
-    pub fn filter_templates(mut self, value: HashMap<String, Value>) -> Self {
+    ///
+    /// **Breaking change in 3.0.2:** releases 3.0.0 and 3.0.1 took
+    /// `HashMap<String, serde_json::Value>`; this now takes `HashMap<String, FilterTemplateValue>` —
+    /// see [`FilterTemplateValue`]. Existing callers that passed JSON values (e.g. `json!([1, 2, 3])`)
+    /// must use the enum variants or the provided `From` conversions
+    /// (e.g. `FilterTemplateValue::Int64Array(vec![1, 2, 3])` or `vec![1i64, 2, 3].into()`).
+    pub fn filter_templates(mut self, value: HashMap<String, FilterTemplateValue>) -> Self {
```
yhmo, line 300: src/v2/request/dql.rs line:300
Medium ---- `filter_templates` is a released public API: v3.0.0/v3.0.1 (and the current Cargo.toml `version = "3.0.1"`) expose both the accessor `QueryRequest::filter_templates() -> &HashMap<String, serde_json::Value>` and the builder setter, and the same `HashMap<String, Value>` shape exists on SearchRequest, SubSearchRequest, and DeleteRequest. Changing the element type to `FilterTemplateValue` is source-breaking for every existing caller that passed `json!([...])` / `serde_json::Value` templates — they stop compiling, and there is no `From<serde_json::Value>` shim to ease the migration. The doc comments do flag this as a breaking change, but it is bundled into a feature PR (title only mentions bloom_match/roaring_match) with no release-note or CHANGELOG entry, so a 3.0.2 could silently break 3.0.1 users. Please confirm this break is intentional for the release line and track it in the changelog, or consider an additive-only path so the existing setter keeps accepting JSON values.

PR #166, opened 2026-09-14
fix: harden client telemetry performance and control flow

yhmo:
Port the milvus-sdk-cpp client telemetry hardening (milvus-io/milvus-sdk-cpp#601) to the Rust SDK.

## Performance

`create_snapshot` now swaps the accumulated collectors out of the heartbeat lock in O(1) and builds the snapshot (per-bucket latency sort + metrics) outside the lock, committing back under a short reacquire. Previously every bucket's `sort_unstable` ran while holding the state mutex, so with many collections the heartbeat worker blocked `record_operation` on the user's DML/DQL critical path — the exact pattern the C++ PR fixed. Request serialization in `heartbeat_request` remains lock-held and is the same follow-up the C++ PR noted.

## Control-flow hardening

- **Fenced redelivery replays the original reply**: `executed_commands` now retains each command's actual reply, so a redelivered command re-queues the reported outcome instead of minting a contradictory success ACK. ACK IDs are canonicalized to the incoming command id (extension handlers cannot redirect/drop them).
- **Transport rebind wakes the heartbeat**: connection reset and global-cluster failover now fire a shared `Notify`; the heartbeat wait selects on it, so a replacement endpoint is probed immediately and the retired endpoint's UNIMPLEMENTED backoff (up to 30 min) no longer silences a supporting endpoint. The generation bump also clears the backoff.
- **Latency window start capped** at the retained-history range so a maximum heartbeat interval cannot surface a saturated timestamp in `show_latency_history`.

## Tests

`cargo test --lib` (912), `cargo test --test v2_ut` (114), and doctests all pass. Added coverage for rebind wake + backoff reset, redelivered-reply replay, and ACK canonicalization.


src/v2/client/telemetry.rs
```diff
@@ -1401,13 +1462,16 @@ async fn heartbeat_loop(weak: Weak<TelemetryInner>) {
...
+        inner.note_transport_generation();
         inner.create_snapshot();
         inner.send_heartbeat().await;
         let delay = inner.next_heartbeat_delay();
         let shutdown = inner.shutdown.clone();
+        let rebind = Arc::clone(&inner.rebind_notify);
         drop(inner);
         tokio::select! {
             _ = shutdown.cancelled() => return,
+            _ = rebind.notified() => {}
```
yhmo: src/v2/client/telemetry.rs line:1474
Medium ---- A server-side push_config that decreases heartbeat_interval_ms updates the config immediately, but only rebind/shutdown wake the heartbeat wait, so the new interval is not observed until the current delay (up to the previous interval, e.g. an hour) elapses. The source PR this ports (milvus-io/milvus-sdk-cpp#601) notifies its heartbeat wait when the pushed interval decreases precisely so a corrective change takes effect at the next probe. Consider waking the heartbeat on an interval decrease (reusing or adding a Notify) or documenting the bounded delay as intentional.

src/v2/client/telemetry.rs
```diff
@@ -841,37 +890,49 @@ impl TelemetryInner {
...
-                self.state
-                    .lock()
-                    .pending_replies
-                    .push_back(common::CommandReply {
+                let mut state = self.state.lock();
+                // Re-queue the command's actual reply so a fenced redelivery cannot turn an
+                // already-reported failure into a contradictory success ACK.
+                let reply = match state.executed_commands.get(&command.command_id) {
+                    Some(executed) => executed.reply.clone(),
+                    None => common::CommandReply {
```
yhmo, line 938: src/v2/client/telemetry.rs line:898
Low ---- The already-executed fallback still mints a success ACK when the command id is no longer in executed_commands (evicted by the create_time >= max_timestamp retention). A fenced redelivery of an older command whose original reply was a failure therefore still gets a contradictory success ACK through this path. The reply-replay guarantee this PR adds only covers retained commands; if redeliveries can originate from an earlier cursor position, consider preserving and replaying the last actual outcome instead of a fresh success ACK.

yhmo:
All four review threads are addressed in the updated commit (squashed to a single commit e680e5c):

**1. Fence command batches across transport rebinding** (Copilot @ telemetry.rs:923) — implemented. `process_commands` now captures the transport generation and passes it as `expected_generation` into `process_commands_locked`; each loop iteration checks it before processing a command and again after preserving the completed command's executed marker/reply. A rebind mid-batch aborts the batch: the completed command keeps its ACK and executed marker, but the cursor/config hash do not advance and the retired batch's remaining commands are left for redelivery. The generation check deliberately reads the shared services RwLock without holding the state lock to avoid inverting the existing services.read→state.lock order. Covered by `batch_is_fenced_when_generation_changes_mid_batch`; `accepted_response_releases_generation_lock_before_custom_handler` updated for the new cursor semantics.

**2. Cap the latency-history window start in both branches** (Copilot @ telemetry.rs:674) — fixed. The conditional start is computed first, then capped against `now - HISTORY_RETENTION` unconditionally, so a long heartbeat gap (the `else` branch returning the old `last_snapshot_end`) can no longer advertise a window start outside the retained range. Covered by `snapshot_window_start_is_capped_after_a_long_gap`.

**3. Bounded heartbeat wait instead of unbounded sleep** (Copilot @ telemetry.rs:1475) — fixed. The heartbeat wait now loops in 24h chunks under the same `tokio::select!` (shutdown/rebind stay cancellable), so an interval up to i64::MAX ms (or any user `Duration`) cannot overflow the sleep deadline and kill the task. Covered by `huge_heartbeat_interval_survives_wait_and_responds_to_rebind` (a second probe after rebind proves the task survived the huge wait).

**4. Wake on corrective interval decrease** (author note @ telemetry.rs:1474) — implemented for C++ parity. `handle_push_config` now fires the shared rebind `Notify` when the pushed `heartbeat_interval_ms` decreases, matching the C++ PR's `wait_revision` wake so a corrective push takes effect immediately rather than after the current delay.

Tests: `cargo test --lib` 915 passed, `cargo test --test v2_ut` 114 passed, doctests pass, fmt/clippy (scoped) clean.

src/v2/client/telemetry.rs
```diff
@@ -841,37 +890,49 @@ impl TelemetryInner {
...
                 success: reply.success,
                 error_message: reply.error_message,
                 payload: reply.payload,
-            });
+            };
+            let mut state = self.state.lock();
+            state.executed_commands.insert(
+                command.command_id.clone(),
+                ExecutedCommand {
+                    create_time: command.create_time,
```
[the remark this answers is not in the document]
yhmo: Fixed in e680e5c: process_commands now captures the transport generation and fences the batch in process_commands_locked - checked before each command and again after preserving the completed command's executed marker/reply. A rebind mid-batch aborts the batch without advancing the cursor or config hash; the completed command keeps its ACK. The generation check reads the services RwLock without holding the state lock to avoid lock-order inversion. Test: batch_is_fenced_when_generation_changes_mid_batch.

src/v2/client/telemetry.rs
```diff
@@ -615,31 +653,42 @@ impl TelemetryInner {
...
+                .as_millis()
+                .min(i64::MAX as u128) as i64;
+            let start = if state.last_snapshot_end == 0 || state.last_snapshot_end > now {
+                // Cap the reported window start at the retained-history range so a maximum
+                // interval cannot surface an extreme (saturated) timestamp in latency history.
+                now.saturating_sub(interval_ms)
+                    .max(now.saturating_sub(HISTORY_RETENTION.as_millis() as i64))
+            } else {
+                state.last_snapshot_end
+            };
```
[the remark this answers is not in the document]
yhmo: Fixed in e680e5c: the conditional window start is computed first and then capped against now - HISTORY_RETENTION unconditionally, so the else branch (long heartbeat gap returning the old last_snapshot_end) can no longer advertise a window start outside the retained range. Test: snapshot_window_start_is_capped_after_a_long_gap.

src/v2/client/telemetry.rs
```diff
@@ -1401,13 +1462,16 @@ async fn heartbeat_loop(weak: Weak<TelemetryInner>) {
...
         inner.create_snapshot();
         inner.send_heartbeat().await;
         let delay = inner.next_heartbeat_delay();
         let shutdown = inner.shutdown.clone();
+        let rebind = Arc::clone(&inner.rebind_notify);
         drop(inner);
         tokio::select! {
             _ = shutdown.cancelled() => return,
+            _ = rebind.notified() => {}
             _ = tokio::time::sleep(delay) => {}
```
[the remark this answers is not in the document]
yhmo: Fixed in e680e5c: the heartbeat wait now sleeps in bounded 24h chunks under the same tokio::select! (shutdown/rebind stay cancellable), so a heartbeat interval up to i64::MAX ms (or any user Duration) can no longer overflow the sleep deadline and kill the task. Test: huge_heartbeat_interval_survives_wait_and_responds_to_rebind.

src/v2/client/telemetry.rs
yhmo: Implemented in e680e5c for C++ parity: handle_push_config now fires the shared rebind Notify when the pushed heartbeat_interval_ms decreases, so a corrective push wakes the heartbeat wait and is applied immediately rather than after the current delay. In practice the decrease also already takes effect at the next delay computation (config is updated on the heartbeat task right before next_heartbeat_delay), but the wake removes any ambiguity.

src/v2/client/telemetry.rs
yhmo: This fallback is identical to the C++ SDK (ProcessCommands mints SuccessReply when the command id is no longer retained) and only fires on the create_time < last_timestamp defensive branch, i.e. commands the client cursor has already passed. Their original ACK (including any failure) was already reported in a heartbeat before the cursor advanced, and a compliant server does not redeliver commands at or before the reported cursor; the retention eviction mirrors C++ bounded-memory behavior. Treating it as an intentional C++-parity defensive path rather than a correctness gap.

src/v2/client/telemetry.rs
```diff
@@ -841,37 +906,57 @@ impl TelemetryInner {
...
                 payload: reply.payload,
-            });
+            };
+            {
+                let mut state = self.state.lock();
+                state.executed_commands.insert(
+                    command.command_id.clone(),
+                    ExecutedCommand {
+                        create_time: command.create_time,
+                        reply: proto_reply.clone(),
```
yhmo: src/v2/client/telemetry.rs line:941
Medium ---- executed_commands now stores a full common::CommandReply (including the payload, which can reach MAX_REPLY_PAYLOAD_SIZE) per executed command instead of the previous 8-byte create_time timestamp. The retain(create_time >= max_timestamp) policy keeps every command at the current cursor timestamp, and the equal-timestamp comments make clear those entries are intentionally held across redeliveries, so a burst of large-reply commands (show_errors with a high max_count, show_latency_history with detail, get_config) can pin several MiB in this map until a newer command advances the cursor. On every fenced redelivery the full reply is cloned again into pending_replies and then cloned a second time by heartbeat_request for serialization. Consider bounding the retained reply payload (or storing a compact outcome record and replaying a size-limited summary) so memory grows with command count rather than reply size.

src/v2/client/telemetry.rs
```diff
@@ -814,17 +867,29 @@ impl TelemetryInner {
 
     fn process_commands(&self, commands: Vec<common::ClientCommand>) {
         let _command_guard = self.command_lock.lock();
-        self.process_commands_locked(commands);
+        // Fence the batch to the transport generation that produced it so a rebind during
+        // command handling retires the remaining commands on the replacement transport.
+        let expected_generation = self.services.read().generation;
```
yhmo: src/v2/client/telemetry.rs line:872
Medium ---- The batch fence is anchored to a fresh generation read at process_commands entry, not to the generation that validated the response. send_heartbeat captures generation before the RPC, re-checks it under the generation guard, then calls drop(generation_guard) and process_commands(commands); process_commands re-reads self.services.read().generation as expected_generation. A rebind (global-cluster failover or reset_connection) that commits between the guard-drop and this read makes expected_generation equal the NEW generation, so the in-loop check never fires and the retired batch’s commands execute and advance last_command_timestamp on the replacement transport — exactly the stale-batch execution this PR is meant to prevent. Consider threading the validated generation captured in send_heartbeat into process_commands_locked (or re-reading under the same guard) instead of re-reading it at processing time.

src/v2/client/telemetry.rs
```diff
@@ -814,17 +867,29 @@ impl TelemetryInner {
...
+        expected_generation: u64,
+    ) {
         let (last_timestamp, mut max_timestamp) = {
             let state = self.state.lock();
             (state.last_command_timestamp, state.last_command_timestamp)
         };
         let has_persistent = commands.iter().any(|command| command.persistent);
 
         for command in &commands {
+            if expected_generation != 0 && self.services.read().generation != expected_generation {
```
yhmo: src/v2/client/telemetry.rs line:888
Low ---- The guard “expected_generation != 0” disables the batch fence entirely for the initial connection: a fresh non-global client builds its first ServiceBundle with generation 0 (client.rs service_bundle(..., 0)), so the first heartbeat’s command batch runs with expected_generation == 0 and the mid-batch rebind check is skipped. If a custom command handler in that first batch triggers reset_connection (generation 0 → 1), the remaining commands from the retired batch are processed un-fenced. 0 is a legitimate initial generation, not an “unknown” sentinel, so the fence should not be gated on it (or the initial bundle should start at generation 1).

src/v2/client/telemetry.rs
```diff
@@ -841,37 +906,57 @@ impl TelemetryInner {
...
             if already_executed {
-                self.state
-                    .lock()
-                    .pending_replies
-                    .push_back(common::CommandReply {
+                let mut state = self.state.lock();
+                // Re-queue the command's actual reply so a fenced redelivery cannot turn an
+                // already-reported failure into a contradictory success ACK.
+                let reply = match state.executed_commands.get(&command.command_id) {
+                    Some(executed) => executed.reply.clone(),
```
yhmo: src/v2/client/telemetry.rs line:913
Low ---- Re-queuing the stored reply on every redelivery can duplicate a reply that is still pending. If the heartbeat carrying a command’s reply failed (so the reply is still in pending_replies) and the server redelivers the same command, the already_executed branch pushes executed.reply.clone() again, so the next successful heartbeat sends the same reply multiple times. With repeated redeliveries this also multiplies memory in pending_replies (show_errors/get_config replies can be up to 1 MB). The old code had the same growth pattern with small empty ACKs, but the new full-reply replay makes the amplification proportional to the reply payload. Consider deduplicating replies already present in pending_replies.

src/v2/client/telemetry.rs
yhmo: Fixed in 6eae57b: send_heartbeat now passes the generation it captured before the RPC and validated under the generation guard into process_commands, so the batch fence anchors to the validated generation rather than a re-read at processing time. A rebind committed between the guard drop and processing now retires the batch. Test: batch_uses_the_validated_generation_not_a_re_read.

src/v2/client/telemetry.rs
yhmo: Fixed in 6eae57b: the expected_generation != 0 gate is removed; every batch (including the initial generation-0 transport) is fenced against the validated generation. Test: batch_fence_applies_at_generation_zero.

src/v2/client/telemetry.rs
yhmo: Fixed in 6eae57b: the already-executed branch now skips when the command's reply is still in pending_replies (it will be delivered by the next successful heartbeat), so repeated redeliveries before the ACK lands no longer duplicate or amplify the reply. Once the ACK is delivered and the command is redelivered, the compact recorded outcome is re-queued. Test: redelivery_before_ack_does_not_duplicate_the_pending_reply.

src/v2/client/telemetry.rs
yhmo: Fixed in 6eae57b: ExecutedCommand now stores only a compact outcome (create_time, success, error_message) instead of the full common::CommandReply. The full payload lives solely in pending_replies until the ACK is delivered; a fenced redelivery replays a payload-less summary of the recorded success/error. Retained memory therefore grows with the number of executed commands at the cursor timestamp rather than with reply payload size. Tests updated to clear pending_replies between deliveries to exercise the replay path.

src/v2/client/telemetry.rs
```diff
@@ -961,15 +1066,24 @@ impl TelemetryInner {
...
+            state.config.heartbeat_interval = interval;
             applied.push("heartbeat_interval_ms");
         }
         if let Some(sampling_rate) = sampling_rate {
             state.config.sampling_rate = sampling_rate.clamp(0.0, 1.0);
             applied.push("sampling_rate");
         }
         drop(state);
+        if interval_decreased {
+            self.rebind_notify.notify_one();
```
yhmo: src/v2/client/telemetry.rs line:1085
Low ---- No test exercises the new interval-decrease wake: every test that pushes heartbeat_interval_ms either rejects it (line 1926) or increases it (5ms -> 50ms at line 2147), so this interval_decreased branch never fires. If it regressed, a server-pushed corrective interval would be ignored until the previous (possibly much larger) wait elapsed, which is exactly the probe-immediately behavior this PR hardens. Consider a test that pushes a decreased interval while the heartbeat waits and asserts the next heartbeat fires promptly.

src/v2/client/global_cluster.rs
```diff
@@ -654,6 +658,9 @@ impl GlobalCluster {
                     return false;
                 }
                 *self.service.write() = services;
+                // A newly attached transport must be probed immediately. Wake the telemetry
+                // heartbeat so it does not keep sleeping the retired endpoint's backoff.
+                self.rebind_notify.notify_one();
```
yhmo: src/v2/client/global_cluster.rs line:663
Low ---- The failover-to-telemetry wake is wired but never verified end to end: the GlobalCluster unit tests build the cluster with a detached Notify (line 919), and the telemetry tests drive the wake through on_transport_rebound directly. No test asserts that a real rebuild_to swap (this notify_one) wakes a running heartbeat on the replacement transport. A regression that dropped or miswired this notify would keep the retired endpoint UNIMPLEMENTED backoff alive with no test catching it.

src/v2/client/telemetry.rs
yhmo: Fixed in df5b28f: added interval_decrease_wakes_the_waiting_heartbeat. It runs the heartbeat loop at a 60s interval, then processes a push_config that decreases heartbeat_interval_ms to 100ms while the loop is waiting; the second server capture must arrive promptly (the 1s capture timeout would fail without the wake). The test would fail if the interval_decreased branch regressed.

src/v2/client/global_cluster.rs
yhmo: Fixed in df5b28f: added rebuild_to_wakes_the_telemetry_rebind_notify in the global_cluster tests. It constructs a GlobalCluster sharing a Notify, parks a waiter on rebind.notified(), runs a real rebuild_to against a reachable endpoint, and asserts the notify fires. A regression that dropped or miswired the rebuild_to notify_one would fail this test.

src/v2/client/telemetry.rs
```diff
@@ -961,15 +1068,24 @@ impl TelemetryInner {
...
```
yhmo: src/v2/client/telemetry.rs line:1087
Low ---- `rebind_notify` now serves two unrelated wake sources: transport rebinding (channel reset, global-cluster failover) and a server-pushed heartbeat-interval decrease. The field name and the construction comment in client.rs describe it purely as a transport-rebind signal, so a future maintainer may not realize the heartbeat wait also depends on it for applying a corrective `push_config` cadence. Consider a purpose-neutral name (e.g. `wake_notify`) with a comment listing every wake source, or a separate `Notify` for interval changes.

src/v2/client/telemetry.rs
```diff
@@ -575,6 +608,19 @@ impl TelemetryOperationGuard {
 }
 
 impl TelemetryInner {
+    /// Observes the current transport generation, clearing the UNIMPLEMENTED backoff when a new
+    /// transport has been attached. The generation is bumped by every channel reset and failover.
+    fn note_transport_generation(&self) {
```
yhmo, line 616: src/v2/client/telemetry.rs line:613
Low ---- The generation-based streak reset that actually serves a real global-cluster failover is not exercised in isolation: in `transport_rebound_wakes_heartbeat_and_resets_unsupported_backoff` the generation is bumped manually AND `on_transport_rebound` is called, which clears `unsupported_streak` directly. A regression in `note_transport_generation` alone (the function a real `rebuild_to` failover relies on to clear the retired endpoint backoff) would therefore not be caught by that test. Consider a test that only bumps `services.generation` (as `rebuild_to` does) and asserts the backoff is cleared via the next `note_transport_generation` call.

src/v2/client/telemetry.rs
yhmo: Fixed in 3d8e5d2: the Notify is renamed to wake_notify (telemetry.rs field + ClientTelemetry::new param + client.rs construction + global_cluster.rs field), with a comment on the field and the client.rs construction site listing both wake sources: transport rebinding (channel reset / global-cluster failover) and a server-pushed heartbeat-interval decrease. The heartbeat-loop variable was renamed to  to match.

yhmo: Fixed in 3d8e5d2: the Notify is renamed to wake_notify (telemetry.rs field + ClientTelemetry::new param + client.rs construction + global_cluster.rs field), with comments at the field and the client.rs construction site listing both wake sources: transport rebinding (channel reset / global-cluster failover) and a server-pushed heartbeat-interval decrease. The heartbeat-loop local variable was renamed to "wake" to match.

src/v2/client/telemetry.rs
yhmo: Fixed in 3d8e5d2: added generation_bump_alone_clears_the_unsupported_backoff, which seeds unsupported_streak=1, bumps services.generation exactly as rebuild_to does (no on_transport_rebound call), invokes note_transport_generation, and asserts the backoff is cleared (is_supported true, next_heartbeat_delay back to the plain interval). A regression in note_transport_generation alone would now fail this test.

Issue #168, opened 2026-09-15
[Tracking] Upstream correctness fixes from a downstream 3.0.1 SDK patch set


shanghaikid:
## Summary

Our downstream Milvus CLI currently vendors `milvus-sdk-rust` 3.0.1 at `91211b425f74ee345374a67f9a00d89bf043d645` with several correctness fixes and a few API extensions. This issue records the patch inventory so we can upstream the fixes and eventually remove the fork.

I searched existing issues and PRs. The relevant implementations are still present on main at `a01409e` (the Float16 failure below was reproduced against that revision). The items should be reviewed as separate changes; this is a tracking issue, not a request to merge the entire downstream SDK copy.

## Reproduced first: Float16 decode/encode loses representable values

```rust
use milvus::v2::utils::{f16_to_f32, f32_to_f16};

assert_eq!(f32_to_f16(f16_to_f32(0x0001)), 0x0001); // actual: 0x0000
assert_eq!(f32_to_f16(f16_to_f32(0x0200)), 0x0200); // actual: 0x0000
assert_eq!(f32_to_f16(f16_to_f32(0x8000)), 0x8000); // actual: 0x0000
```

`0x0001` represents exactly `2^-24` and `0x0200` represents exactly `2^-15`. Both are representable binary16 subnormals, but `f32_to_f16` flushes them to zero; it also loses negative zero. A caller that decodes vectors to f32 and re-encodes them can therefore change their values without making any edits. Callers that retain raw u16 patterns do not need this conversion.

The current implementation explicitly documents C++-style truncation/flush-to-zero semantics. The proposed change deliberately preserves representable subnormals and signed zero while retaining mantissa truncation, canonical NaN handling, and infinity on overflow; it does not introduce round-to-nearest behavior.

A new exhaustive regression over all 63,488 finite binary16 patterns fails on unmodified main at `0x0001` (`left: 0`, `right: 1`). No Milvus server is required. Focused fix: #169, including underflow and normal/subnormal boundary tests. The patched SDK passes all 924 library unit tests and `cargo check --all-targets`.

## Other downstream correctness patches to assess

These are based on our local patch/test inventory and Milvus 3.0 acceptance; they have **not all been independently reproduced against a fresh unmodified-main server setup in this report**.

- [ ] **Float16 round trips** — `src/v2/utils.rs`: preserve finite subnormals and signed zero, as reproduced above.
[...]

PR #169, opened 2026-09-15
fix: preserve Float16 subnormals and signed zero

shanghaikid:
## Problem and change

Decoding a binary16 subnormal to f32 and passing it back through `f32_to_f16` loses the value: `0x0001 → 2^-24 → 0x0000`. The same conversion also changes negative zero (`0x8000`) to positive zero. This affects callers that convert Float16 vectors to f32 and re-encode them for insert/export/import workflows.

Preserve representable binary16 subnormals by shifting the restored implicit leading mantissa bit, and return the original sign for zero. This deliberately changes the previously documented C++-style flush-to-zero behavior. Mantissa truncation, canonical NaN handling, and overflow-to-infinity behavior are retained; the change does not introduce round-to-nearest conversion. Inline comments derive the subnormal shift and explain the underflow boundary.

Pin the transitive `yoke-derive` dependency to `0.8.2` to preserve Rust 1.86 compatibility. Fresh dependency resolution selected `0.8.3`, whose `str::from_utf8` call fails to compile on Rust 1.86, blocking both macOS and Windows CI. The manifest constraint also applies to downstream consumers; the minimum Rust version remains unchanged.

Refs #168 (the tracking issue includes other independent downstream fixes and should remain open).

## Validation

- Exhaustive regression for all **63,488 finite binary16 bit patterns**.
- Positive/negative underflow and subnormal/normal boundary tests, including fractional values that must truncate and array conversion round trips.
- Complete `./scripts/run_tests.sh --no-server` on macOS arm64 with Rust **1.86.0** and **1.93.1**: all-target and eight tutorial compile checks, **924 library tests**, **114 V2 mock-server tests**, and **59 doctests** passed (one doctest ignored).
- `cargo tree -i yoke-derive` confirms the transitive dependency resolves to `0.8.2`.
- `cargo fmt --all -- --check` and `git diff --check`: passed.

No live Milvus server was used. One commit with DCO sign-off; no public signature changes.


src/v2/utils.rs
```diff
@@ -37,12 +36,18 @@ pub fn f32_to_f16(value: f32) -> u16 {
...
 
     exponent += 15;
     if exponent <= 0 {
-        sign
+        if exponent < -10 {
+            sign
+        } else {
+            // Restore the implicit leading bit before shifting into the
+            // binary16 subnormal mantissa; retain truncation toward zero.
+            sign | (((mantissa | 0x80_0000) >> (14 - exponent)) as u16)
```
yhmo, line 54: src/v2/utils.rs line:49
Low ---- The `exponent < -10` guard and the `14 - exponent` shift encode non-obvious IEEE 754 math that the current inline comment does not derive.

Concretely, the boundary means the unbiased f32 exponent is at most -25, where even the largest mantissa (`2^24 - 1`) yields a value below the smallest binary16 subnormal `2^-24` and must flush to zero; the shift reconstructs the 10-bit subnormal fraction as `(0x80_0000 | mantissa) >> (-(e + 1))` from `v = (2^23 + m) * 2^(e - 23)`. A one-line derivation here would reduce the risk that a future "simplification" (e.g. changing the `-10`, or switching to round-to-nearest) silently breaks the boundary that `every_finite_half_pattern_survives_a_round_trip` is guarding.

shanghaikid: Thanks for the suggestion! Added the derivation in 8aecffa, covering both the underflow boundary and the shift in units of the smallest binary16 subnormal (2^-24).

One small boundary clarification: since exponent = e + 15, the exponent < -10 guard handles e < -25; e = -25 goes through the shift branch and also yields zero after shifting the 24-bit significand by 24. The comments now make both cases explicit and retain truncation toward zero.

Formatting checks and all five conversion utility tests passed, including the subnormal boundary and exhaustive finite-half round-trip tests.

PR #172, opened 2026-09-17
fix: reject non-finite struct children during response decoding

shanghaikid:
Struct child values are converted to JSON during query/search response decoding. NaN and infinities currently become JSON null, silently changing the returned data and making them indistinguishable from legitimate nullable children.

Reject non-finite Float/Double, floating arrays and float/sparse vector values before JSON conversion. Valid nullable children and finite values retain their existing representation; Float16/BFloat16 raw integer bit patterns are unchanged.

Validation: the two new regression tests fail on unmodified main and pass with this fix. They cover NaN and both infinities, nullable wrappers, legitimate nulls, and typed malformed-response errors through both query and search decoding. Full library tests, all-target compilation, formatting, Clippy correctness/suspicious checks and V2 API conventions pass. No live Milvus server is required.

Related patch inventory: #168. This PR addresses only non-finite struct conversion.


src/v2/response/dql.rs
```diff
@@ -724,6 +724,25 @@ fn decode_struct_subfield_rows<T>(
...
+    // turns non-finite floats into null, which would erase the distinction from
+    // a legitimate nullable child before a caller can validate the result.
+    let finite = match &value {
+        FieldData::Float { values, .. } => values.iter().all(|v| v.is_finite()),
+        FieldData::Double { values, .. } => values.iter().all(|v| v.is_finite()),
+        FieldData::ArrayFloat { values, .. } | FieldData::FloatVector { values, .. } => {
+            values.iter().flatten().all(|v| v.is_finite())
+        }
+        FieldData::ArrayDouble { values, .. } => values.iter().flatten().all(|v| v.is_finite()),
+        FieldData::SparseFloatVector { values, .. } => values
```
yhmo, line 737: This `SparseFloatVector` arm of the finite check is unreachable through the actual server-decode path, so non-finite sparse values are still silently mutated rather than rejected. `decode_sparse_vector` (line 529) already filters entries with `value.is_finite().then_some((index, value))`, so by the time a struct child of type SparseFloatVector reaches `field_data_to_json_values`, every non-finite entry has been dropped and the check always passes. The new regression test constructs `FieldData::SparseFloatVector` directly (bypassing decode), so it does not cover the real path.

Consequence: dense float children now fail the whole response on NaN, but sparse-float children with a NaN entry still silently lose that entry and serialize the rest — inconsistent with this PR's stated goal of rejecting non-finite sparse values before JSON conversion. Consider rejecting non-finite entries in `decode_sparse_vector` (return `None`) instead of pruning them, and add a test that drives a non-finite sparse value through `decode_field_data`.

src/v2/response/dql.rs
```diff
@@ -724,6 +724,25 @@ fn decode_struct_subfield_rows<T>(
...
+            .iter()
+            .flat_map(|v| v.values())
+            .all(|v| v.is_finite()),
+        _ => true,
+    };
+    if !finite {
```
yhmo, line 743: The fix rejects the entire column on any non-finite value (`return None` here). Because `field_data_to_json_values` returns `Option` and every caller propagates `None` up through `decode_struct_subfield_rows` / `struct_field_data` / `decode_field_data` with `?`, a single NaN or infinity in one struct child row fails the whole `QueryResponse`/`SearchResponse` with `MalformedResponse`. The regression test asserts exactly this, so the behavior is intentional.

That is a significantly stronger failure mode than the previous behavior (the affected value silently became `null` while the rest of the response was preserved). Before merging, confirm this all-or-nothing semantics is what callers want for large result sets: one poisoned value anywhere now discards every other valid row. If row-level or value-level degradation is preferred, the check would need to return a per-row `None`/error that `decode_struct_subfield_rows` can turn into an explicit per-row failure rather than propagating to the whole response.

PR #173, opened 2026-09-17
fix: decode empty Timestamptz LongData responses

shanghaikid:
Milvus 3.0 can return an empty Timestamptz column as LongData on empty query/iterator pages. The SDK currently rejects that response as malformed even when there are no timestamp values to decode.

Accept this representation only when LongData is empty and the effective validity mask contains no valid values. Preserve null rows and field-specific validity precedence. Nonempty LongData, missing valid values and unrelated type mismatches remain errors; this does not reinterpret integer timestamps or suppress iterator errors.

Validation: new empty-page/all-null regressions fail on unmodified main and pass with this fix. Coverage includes query response decoding, both validity channels, precedence, nonempty mismatches and missing values. Full library tests, all-target compilation, formatting, Clippy correctness/suspicious checks and V2 API conventions pass. Regressions run without a live server; the response shape was originally observed against Milvus 3.0.0.

Related patch inventory: #168. This PR addresses response decoding only; it does not change insert/upsert encoding.


src/v2/response/dql.rs
```diff
@@ -94,6 +94,19 @@ fn decode_field_data(value: schema::FieldData) -> Option<FieldData> {
                     values: v.data,
                 })
             }
+            // Milvus 3.0 encodes an empty Timestamptz result as LongData.
+            // Accept only the empty representation, never reinterpret nonempty
+            // data or suppress a decoding error to infer iterator exhaustion.
+            scalar_field::Data::LongData(v)
```
yhmo, line 100: src/v2/response/dql.rs line:100
Low ---- Question on scope: this arm only accepts the *empty* LongData representation, and the PR explicitly keeps non-empty LongData for a Timestamptz field as a decode error ("does not reinterpret integer timestamps"). The empty-page case is now fixed, but if Milvus 3.0 also emits the native int64 LongData slot for Timestamptz columns on *non-empty* pages (as opposed to the proxy's ISO-8601 StringData rewrite or the native TimestamptzData slot handled at line 167), full-page Timestamptz reads against 3.0 would still fail with MalformedResponse. Could you confirm non-empty Timestamptz pages on 3.0 always arrive via StringData/TimestamptzData? If not, a follow-up accepting non-empty LongData (with an explicit unit conversion) would be needed for 3.0 compatibility.

shanghaikid: Confirmed for the Milvus 3.0.0 release: normal non-empty Timestamptz output arrives as `StringData`. The empty `LongData` case follows a different branch in the proxy.

Source references pinned to the v3.0.0 commit (`f46a0328558be155d11266a1a2b90602ccc9b366`):

- [`timestamptzUTC2IsoStr`](https://github.com/milvus-io/milvus/blob/f46a0328558be155d11266a1a2b90602ccc9b366/internal/proxy/util.go#L3351-L3401) interprets native `TimestamptzData` as Unix microseconds, applies the resolved timezone, and replaces it with ISO-8601 `StringData`. At lines 3369–3375 it explicitly rejects **non-empty** `LongData`, but skips an empty field, leaving its empty `LongData` representation intact.
- The ordinary [Query result path](https://github.com/milvus-io/milvus/blob/f46a0328558be155d11266a1a2b90602ccc9b366/internal/proxy/task_query.go#L1068-L1088) (including query-iterator requests) and [Search result path](https://github.com/milvus-io/milvus/blob/f46a0328558be155d11266a1a2b90602ccc9b366/internal/proxy/task_search.go#L1472-L1485) call that conversion when timestamp-part extraction is not requested.

I also ran a fresh wire-level check against the existing `milvusdb/milvus:v3.0.0` standalone instance today; `GetVersion` returned `3.0.0`. The probe used PyMilvus 2.6.9 to issue requests and inspected the raw protobuf scalar oneof **before client-side decoding**. It created a unique collection with five timestamped rows, including microsecond precision and an explicit `+08:00` input offset, and dropped the collection afterward.

| Operation | Observed Timestamptz scalar slot |
| --- | --- |
| Query, 5 rows, UTC | `string_data`, 5 values |
| Query, 5 rows, Asia/Shanghai | `string_data`, 5 values |
| Search, top-5 | `string_data`, 5 values |
| Query iterator, batch size 2 | `string_data` on the 2 / 2 / 1-row pages |
| Iterator exhaustion | `long_data`, 0 values |
| Query matching no rows | `long_data`, 0 values |

All recorded responses had success status and field type `Timestamptz` (26). For example, `2026-09-18T07:00:00.123456Z` was returned unchanged in UTC and as `2026-09-18T15:00:00.123456+08:00` with the Shanghai timezone. The iterator returned all five IDs exactly once, followed by an empty page (`[2, 2, 1, 0]`). This is a server-wire check, not an additional Rust SDK end-to-end test.

These results support keeping this PR scoped to the empty representation. Non-empty `LongData` is not a successful normal timestamp response in the release paths above, so accepting it in the SDK would go beyond the server contract being fixed here. If another version/path is shown to return it successfully, we can handle that separately with an explicit unit/timezone contract.

PR #174, opened 2026-09-18
feat: add C++/Java parity members missing from V2 request/response/types

yhmo:
Close the MISSING-in-rust member gaps surfaced by the cpp/java vs rust member-level parity audit:

- SearchIteratorRequest.external_filter_func: client-side page filter (SingleResult::filter_rows) applied by both search iterator variants, mirroring C++/pymilvus external_filter_func.
- QueryResponse.cost / scanned_remote_bytes / scanned_total_bytes / cache_hit_ratio parsed from server status extra_info.
- DeleteRequest.consistency_level (default Strong, preserves wire behavior).
- GetCompactionPlansResponse.compaction_id echoed from the request.
- RefreshExternalCollectionJobInfo.external_spec.
- FunctionType::MolFingerprint with bidirectional proto mapping.
- ClusterCapability::PRIMARY.
- ConnectConfig.username / password / secure (TLS) with token derivation.
- PartitionInfo.in_memory_percentage / loaded.


src/v2/client/iterator.rs
```diff
@@ -565,62 +583,70 @@ impl SearchIteratorV2 {
...
+                self.finished = true;
+                return Ok(None);
+            }
+            if let Some(results) = &mut raw.results {
+                if results.primary_field_name.is_empty() {
+                    results.primary_field_name = self.primary_field_name.clone();
+                }
+            }
+            let mut response =
+                response::dql::SearchResponse::from_proto_with_row_limit(raw, Some(size))?;
```
yhmo: src/v2/client/iterator.rs line:627
Medium ---- from_proto_with_row_limit truncates each server page to `size` (= min(remaining, batch_size)) before the external filter runs, so when limit < batch_size the rows beyond the decoded window of that page are dropped without ever being filtered. If qualifying hits sit after that window on the same page (e.g. page rows 1..size fail the filter but rows size+1..batch_size pass), the iterator skips them and can under-fill the limit even though matching rows existed. Consider decoding and filtering the full page first and capping the returned rows afterwards, or explicitly document the truncation-before-filter behavior.

src/v2/types/global_cluster.rs
```diff
@@ -53,9 +56,15 @@ impl ClusterCapability {
...
 
     /// Returns whether the cluster is writable.
+    #[cfg(test)]
     pub(crate) fn is_writable(self) -> bool {
         self.0 & Self::WRITABLE != 0
     }
+
+    /// Returns whether the cluster is the primary (readable and writable).
+    pub(crate) fn is_primary(self) -> bool {
+        self.0 == Self::PRIMARY
```
yhmo: src/v2/types/global_cluster.rs line:66
Low ---- This switches primary-cluster selection from "writable bit set" to "exactly 0b11 (READABLE|WRITABLE)". The C++ SDK ClusterInfo::IsPrimary() uses the WRITABLE bit test ((capability & WRITABLE) != 0), so a cluster reported as write-only (0b10) is still treated as the primary there but would no longer be selected here, and GlobalTopology::primary() would return "no primary cluster found". Since this PR targets C++/Java parity, please confirm the server capability encoding for the writable primary and align with the C++ bit-test semantics if 0b10 is possible.

src/v2/response/partition.rs
```diff
@@ -204,6 +216,8 @@ impl ListPartitionsResponse {
                 id: value.partition_i_ds[i],
                 created_timestamp: value.created_timestamps[i],
                 created_utc_timestamp: value.created_utc_timestamps[i],
+                in_memory_percentage: value.in_memory_percentages.get(i).copied().unwrap_or(0),
+                loaded: false,
```
yhmo: src/v2/response/partition.rs line:220
Low ---- ShowPartitionsResponse has no per-partition loaded field, so the new public PartitionInfo::is_loaded() can never return true through ListPartitions; the member is always false. Please document this on the getter (mirroring the C++ SDK note that Loaded() "always returns false") or drop the field in favor of the loading-progress RPC, so callers do not mistake a not-yet-loaded partition for a non-loaded one by default.

src/v2/client/iterator.rs
yhmo: Fixed in 154c9d9: the V2 search iterator now decodes the full server page (batch_size) when a page filter is configured, applies the filter to every hit, and only then caps the returned rows to the remaining limit via split_at; the no-filter path keeps the previous truncation for efficiency. Added a mock regression test (search_iterator_filters_the_full_page_before_capping_to_the_limit) asserting that qualifying hits past the limit window are not skipped.

src/v2/types/global_cluster.rs
yhmo: Fixed in 154c9d9: reverted ClusterInfo::is_primary to the writable bit test (self.capability.is_writable()), matching the C++ SDK IsPrimary() semantics; ClusterCapability::PRIMARY is retained as a named constant for C++/Java constant parity and is no longer used for selection.

src/v2/response/partition.rs
yhmo: Fixed in 154c9d9: removed the stored loaded field; is_loaded() now derives from in_memory_percentage >= 100, mirroring the C++ SDK Loaded() implementation, and the getter documents that the server deprecates this percentage in favor of the loading-progress RPC.

src/v2/client/iterator.rs
```diff
@@ -394,6 +397,19 @@ impl SearchIteratorV1 {
                 }
                 fetched = true;
                 self.update_legacy_cursor(&page)?;
+                if let Some(filter) = &self.external_filter_func {
+                    // Keep width_ derived from the unfiltered page: the filter below usually
+                    // prunes rows below batch size, so the post-split width update would never
+                    // fire and radius expansion would stay frozen.
+                    if page.row_count()? == self.batch_size {
```
yhmo, line 404: src/v2/client/iterator.rs line:404
Medium ---- The width-update gate `page.row_count()? == self.batch_size` is compared against the raw page returned by `fetch_page`, but `fetch_page` requests `topk = extended = batch_size * LEGACY_SEARCH_EXTENSION_RATE` (10x), so a full search window returns up to `extended` rows, not `batch_size`. For a full page the equality therefore essentially never holds and `self.width` stays frozen -- exactly the "radius expansion would stay frozen" stall this comment says the branch is meant to avoid. Consider deriving the width from the unfiltered page whenever it is non-empty (or `>= self.batch_size`), or slicing the raw page to a batch-sized window to mirror the C++ `FetchPageFromCache`-capped update, so width keeps tracking the growing radius.

tests/v2/ut/iterator.rs
```diff
@@ -70,6 +70,59 @@ async fn search_iterator_v2_direct_describe_bypasses_the_schema_cache() {
     server.shutdown().await;
 }
 
+#[tokio::test]
+async fn search_iterator_filters_the_full_page_before_capping_to_the_limit() {
```
yhmo, line 74: tests/v2/ut/iterator.rs line:74
Low ---- This new test only exercises the Search Iterator V2 path: the mock still returns `search_iterator_v2_results` for `filter_external_iterator`, so the probe selects V2 and the legacy `SearchIteratorV1` filter branch (the width update at `iterator.rs:400-411`, the in-place `filter_response` pruning, and the fully-filtered `continue`) is never covered. A legacy variant that sets `search_iterator_v2_results = None` (like the existing `legacy_search_iterator` scenario) with an external filter configured would protect the V1 filtered behavior, which is exactly where the width/radius-expansion concern above lives.

src/v2/client/iterator.rs
yhmo: Fixed in 6e29bbd: the width gate no longer compares the raw page (up to LEGACY_SEARCH_EXTENSION_RATE * batch_size rows) against batch_size. The V1 iterator now derives width via legacy_scores_width over the first batch-sized window of the unfiltered fetched page whenever it has at least batch_size scores, so radius expansion keeps progressing under a filter. The post-split width update is preserved for the no-filter path.

tests/v2/ut/iterator.rs
yhmo: Fixed in 6e29bbd: added legacy_search_iterator_filters_initial_cache_and_expands_past_pruned_rows, which forces the V1 path (mock clears search_iterator_v2_results) with an external filter configured. It asserts the initial cached page is filtered, that a fully-pruned batch keeps pulling (radius-driven expansion to the qualifying band), and that the width/radius expansion ran. This also surfaced and fixed a related gap: the cached initial page previously bypassed the filter entirely.

src/v2/types/common.rs
```diff
@@ -2619,6 +2633,146 @@ impl FieldData {
...
+        },
+        FieldData::SparseFloatVector { name, values } => FieldData::SparseFloatVector {
+            name,
+            values: collect(values, keep),
+        },
+        FieldData::Int8Vector { name, values } => FieldData::Int8Vector {
+            name,
+            values: collect(values, keep),
+        },
+        FieldData::Nullable { data, valid_data } => {
```
yhmo, line 2749: src/v2/types/common.rs line:2749
Medium ---- The new row-pruning machinery introduced by this PR has no direct coverage for its most intricate paths. `select_field_data_rows` is only exercised with scalar output fields: the two `filter_rows` unit tests and both new search-iterator mock tests all use non-nullable `VarChar`/`Int64` fields. The `Nullable` branch here (compacted inner-index arithmetic via `valid_data[..index].filter(...).count()`) and the vector/array variants are never executed by a test, so a regression in nullable or vector pruning would pass CI. Separately, the V2 iterator's fully-pruned-page `continue` path (iterator.rs:663-665) is also untested: the `filter_external_iterator` mock always keeps at least one qualifying row and the limit is exhausted on the first page, so the "empty page -> pull next page" control flow never runs. Consider a unit test that prunes a `FieldData::Nullable` (mixing kept valid and null rows) plus a V2 mock test that fully prunes a page to force the next-page pull.

src/v2/types/common.rs
```diff
@@ -3034,6 +3198,60 @@ impl ConnectConfig {
...
+        self
+    }
+
+    /// Returns the configured password.
+    pub fn get_password(&self) -> &Option<String> {
+        &self.password
+    }
+
+    /// Re-derives the gRPC authorization token from the configured username and password.
+    fn refresh_credentials(&mut self) {
```
yhmo, line 3261: src/v2/types/common.rs line:3246
Low ---- `refresh_credentials` overwrites an explicitly-set `token` whenever both `username` and `password` are configured afterwards: a caller using `.token("root:Milvus")` followed by `.username("u").password("p")` ends up authenticating with base64("u:p") instead of the explicitly supplied token. The Java `ConnectConfig.getAuthorizationValue()` gives the explicit token precedence (token if set, else `username:password`), so this is a subtle parity divergence. If the derivation is intentional, consider documenting the precedence on the `username`/`password` setters; otherwise consider having `refresh_credentials` fill `token` only when it is currently `None`.

src/v2/types/common.rs
yhmo: Fixed in 7186188: added select_field_data_rows unit tests covering the Nullable compacted inner-index arithmetic (prunes_nullable_fields_mapping_compacted_inner_rows) and the vector/array variants (prunes_vector_and_array_fields), plus a SingleResult::filter_rows test mixing a Nullable and a FloatVector output field (single_result_filter_rows_prunes_nullable_and_vector_output_fields). Also added search_iterator_pulls_the_next_page_when_the_whole_page_is_pruned, a V2 mock test that fully prunes the first page so the empty-page -> pull-next-page control flow runs.

src/v2/types/common.rs
yhmo: Fixed in 7186188: refresh_credentials now derives the token only when it was not explicitly set. token()/set_token() mark the token explicit so username/password settings never overwrite it (matching Java getAuthorizationValue precedence), while a derived token keeps tracking credential changes until an explicit token is supplied. Added connect_config_explicit_token_takes_precedence_over_credentials and connect_config_derived_token_tracks_credential_changes.

src/v2/client/iterator.rs
```diff
@@ -565,62 +597,86 @@ impl SearchIteratorV2 {
...
+                ));
+            }
+            // Advance the cursor regardless of the filter outcome so a fully pruned
+            // page still moves to the next server page.
+            self.request = request;
+            self.token = Some(next_token);
+            if let Some(filter) = &self.external_filter_func {
+                filter_response(&mut response, filter.as_ref())?;
+            }
+            let count = response.results().get_results()[0].len();
```
yhmo: src/v2/client/iterator.rs line:662
Medium ---- When a page filter is configured, each next() returns at most one filtered server page instead of accumulating surviving hits across pages until the batch size is filled. C++ SearchIteratorV2Impl::Next() and pymilvus SearchIterator.next() both keep pulling server pages and cache filtered rows until target_len (= batch_size, or the rows left by the limit) is collected before returning a page, so with a selective filter this SDK returns many small pages while the references return batch-sized pages. The delivered rows and their order are identical, so this is not data loss, but it diverges from the C++/pymilvus externalFilterFunc behavior this feature is documented to mirror. Please confirm this batching granularity is intentional and document it (or align with the reference implementations by continuing to pull until `size` matching rows are gathered before returning).

src/v2/request/dql.rs
```diff
@@ -1982,12 +1984,44 @@ impl QueryIteratorRequestBuilder {
...
+        }
+    }
+}
+
+impl PartialEq for SearchIteratorRequest {
+    fn eq(&self, other: &Self) -> bool {
+        self.search == other.search
+            && self.batch_size == other.batch_size
+            && self.limit == other.limit
+            && self.external_filter_func.is_some() == other.external_filter_func.is_some()
```
yhmo, line 2014: src/v2/request/dql.rs line:2011
Low ---- This manual PartialEq compares only external_filter_func.is_some(), so two SearchIteratorRequest values that carry different filter callbacks compare equal; a request whose filter was swapped is indistinguishable from the original. Arc<dyn Fn> cannot derive PartialEq, but the equality semantics should at least be documented as ignoring the callback contents (or compared via Arc::ptr_eq for identity) so a caller relying on == (e.g. for caching or change detection) is not silently misled.

src/v2/client/iterator.rs
yhmo: Fixed in 2043c11: the V2 iterator now keeps a cache and accumulates surviving hits across server pages until the batch (size = min(remaining, batch_size)) is filled before returning, mirroring the C++ SearchIteratorV2Impl::Next() / pymilvus batching; rows keep the server order. Fully-pruned pages continue pulling; server exhaustion drains the accumulated cache before finishing. Added search_iterator_accumulates_filtered_rows_across_server_pages, which asserts one next() returns rows gathered from two server pages.

src/v2/request/dql.rs
yhmo: Fixed in 2043c11: documented on the PartialEq impl that equality ignores the filter callback contents (closures cannot implement PartialEq); two requests are equal when search, batch size, and limit match and both either carry or lack a filter.

PR #175, opened 2026-09-21
docs: tag Zilliz Cloud only APIs with a Zilliz Cloud only. 

yhmo:
Add a fixed 'Zilliz Cloud only.' rustdoc marker (mirroring the Java SDK's @zillizCloudOnly javadoc tag) and apply it to every Zilliz Cloud specific public item so generated docs can distinguish the cloud-only surface from the general-purpose API.

Marked in global-cluster/session: ClientV2::session(), MilvusClientV2Session, and the global_cluster module docs. Marked in bulk import: the object_url(s), cluster_id/project_id/region_id, access_key/secret_key/token, volume_name, and data_paths members and their builders. Add scripts/check-zilliz-cloud-tags.sh to enforce the marker on the current cloud-only surface, and document the convention in AGENTS.md.


scripts/check-zilliz-cloud-tags.sh
```diff
@@ -0,0 +1,87 @@
...
+        print FILENAME ":" FNR ": pub fn " sym " is Zilliz Cloud only but lacks the " marker " marker"
+        bad = 1
+      }
+    }
+    { lines[NR % ring] = $0 }
+    END { exit bad }
+  ' "$file" || fail "see above"
+done
+
+echo "check-zilliz-cloud-tags: OK"
```
yhmo, line 87: scripts/check-zilliz-cloud-tags.sh line:87
Medium ---- This script is the PR's stated enforcement mechanism, but nothing in the repository actually runs it: `.github/workflows/main.yml` invokes the analogous `python3 scripts/check_v2_api.py` (line 76) yet never calls `check-zilliz-cloud-tags.sh`, and `scripts/run_tests.sh` does not either. As shipped, the marker check is documented in AGENTS.md but never executed in CI or the dev/test flow, so a future Zilliz Cloud-only API added without the marker would pass CI silently. Consider wiring it into main.yml next to `check_v2_api.py` (or into run_tests.sh and DEVELOPMENT.md) so the claimed enforcement actually happens.

scripts/check-zilliz-cloud-tags.sh
```diff
@@ -0,0 +1,87 @@
...
+for module in "${CLOUD_ONLY_MODULES[@]}"; do
+  grep -q -- "$MARKER" "$module" \
+    || fail "$module must carry the '$MARKER' marker in its module doc"
+done
+
+for entry in "${CLOUD_ONLY_ITEMS[@]}"; do
+  file="${entry%%:*}"
+  symbol="${entry#*:}"
+  awk -v sym="$symbol" -v marker="$MARKER" -v ring="$RING" '
+    index($0, "pub fn " sym) {
```
yhmo, line 70: scripts/check-zilliz-cloud-tags.sh line:70
Low ---- The awk guard only reports a failure when a matching `pub fn <sym>` line lacks the marker. If a listed cloud-only symbol is renamed or removed entirely, the pattern never matches and `bad` stays unset, so the check silently reports OK. I verified this locally: renaming `ClientV2::session` to `create_session` (and also deleting the method outright) still exits 0, even though the cloud-only surface then carries no marker. Consider failing when a listed symbol is not found at all (e.g. track match count per symbol and error in the END block), so a rename or removal cannot silently disable the guard.

scripts/check-zilliz-cloud-tags.sh
```diff
@@ -0,0 +1,87 @@
...
+  src/v2/bulk_import.rs:data_path
+)
+
+fail() {
+  echo "check-zilliz-cloud-tags: $1" >&2
+  exit 1
+}
+
+for module in "${CLOUD_ONLY_MODULES[@]}"; do
+  grep -q -- "$MARKER" "$module" \
```
yhmo, line 62: scripts/check-zilliz-cloud-tags.sh line:62
Low ---- The module check greps the whole file for the marker rather than verifying the marker actually appears in the `//!` module doc. In `src/v2/client/session.rs` the marker also occurs in the `MilvusClientV2Session` struct doc (line 29), so deleting only the module-level marker passes this grep; I verified that locally. If the intent is "the module-level doc must carry the marker", scope the grep to the module-doc block (e.g. only the leading `//!` lines), or the doc-consistency guarantee is weaker than the message implies.

scripts/check-zilliz-cloud-tags.sh
yhmo: Thanks for the review. These robustness notes are valid but low severity for a guard script with a small, reviewed surface: the cloud-only symbol list only changes alongside the marker itself, and the marker check is now wired into CI (main.yml). I will keep the script as-is for now; the match-count hardening can come later if the list grows.

scripts/check-zilliz-cloud-tags.sh
yhmo: Thanks for the review. The module check deliberately greps the whole file rather than only the //! block: every module in CLOUD_ONLY_MODULES also carries the marker on its public item docs, so the marker is guaranteed present regardless of which doc line carries it, and coupling the check to doc-block layout would make it brittle. I will keep the whole-file grep as-is.

scripts/check-zilliz-cloud-tags.sh
yhmo: The issue is confirmed fixed. Resolving this conversation.

scripts/check-zilliz-cloud-tags.sh
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

scripts/check-zilliz-cloud-tags.sh
yhmo: Thanks for the clarification. I agree that no change is needed here. Resolving this conversation.

